PR #4, opened 2015-04-15
don't swallow sys.exit

Carreau:
back to back with :

https://github.com/ipython/ipython/pull/8324


Carreau:
this lead to some weris behavior with bad jupyterdrive, where a bad traitlet config would raise a sys.exit(1) that would be caught.

PR #7, opened 2015-04-26
WIP: implement migration

### LINKED ISSUE #2: add `jupyter migrate`
explicit command for migrating from an IPython profile to Jupyter

minrk:
Currently, this implements migration of config files and user-installed nbextensions, kernelspecs from IPython < 4.0 to Jupyter locations via `jupyter migrate` as:

- copy files to new locations
- if new locations already exist, do nothing
- never overwrite anything
- don't copy empty config files
- small amount of find-replace in config files for renamed classes (e.g. IPythonQtConsoleApp -> JupyterQtConsoleApp)

To figure out for this PR:

- [x] what files should be migrated
- [x] copy or move, etc.? Allow overwrite?
- [x] API for migration (CLI)
- [x] should there be any modifications to contents, or only copies?

To figure out after this PR:

- When to migrate.

I want to implement migration first, and work on implicit-on-startup logic, if any, as a separate discussion.

cc @takluyver @fperez @ellisonbg

closes #2

### DIFF
--- a/jupyter_core/migrate.py
+++ b/jupyter_core/migrate.py
@@ -64,6 +64,9 @@ config_substitutions = {
 
 def migrate_dir(src, dst):
     """Migrate a directory"""
+    if not os.listdir(src):
+        print("No files in %s" % src)
+        return False
     if os.path.exists(dst):
         if os.listdir(dst):
             # already exists, non-empty
@@ -111,6 +114,12 @@ def migrate_one(src, dst):
 
 
 def migrate_static_custom(src, dst):
+    """Migrate non-empty custom.js,css from src to dst
+    
+    src, dst are 'custom' directories containing custom.{js,css}
+    """
+    migrated = False
+    
     custom_js = pjoin(src, 'custom.js')
     custom_css = pjoin(src, 'custom.css')
     # check if custom_js is empty:
@@ -136,16 +145,20 @@ def migrate_static_custom(src, dst):
     
     if custom_js_empty and custom_css_empty:
         # nothing to migrate
-        return
+        return False
     ensure_dir_exists(dst)
     
     if not custom_js_empty or not custom_css_empty:
         ensure_dir_exists(dst)
     
     if not custom_js_empty:
-        migrate_file(custom_js, pjoin(dst, 'custom.js'))
+        if migrate_file(custom_js, pjoin(dst, 'custom.js')):
+            migrated = True
     if not custom_css_empty:
-        migrate_file(custom_css, pjoin(dst, 'custom.css'))
+        if migrate_file(custom_css, pjoin(dst, 'custom.css')):
+            migrated = True
+    
+    return migrated
 
 
 def migrate_config(name, env):
--- a/jupyter_core/migrate.py
+++ b/jupyter_core/migrate.py
@@ -172,7 +172,7 @@ def migrate():
         'jupyter_data': jupyter_data_dir(),
         'jupyter_config': jupyter_config_dir(),
         'ipython_dir': get_ipython_dir(),
-        'profile': locate_profile(),
+        'profile': os.path.join(get_ipython_dir(), 'profile_default'),
     }
     for src_t, dst_t in migrations.items():
         src = src_t.format(**env)
--- a/jupyter_core/migrate.py
+++ b/jupyter_core/migrate.py
@@ -91,7 +91,7 @@ def migrate_file(src, dst, substitutions=None):
         return False
     print("Copying %s -> %s" % (src, dst))
     ensure_dir_exists(os.path.dirname(dst))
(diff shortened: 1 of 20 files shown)


Carreau:
> copy files to new locations

I would move it, but copy is fine. 
The reason to move is that if file still exist in old location, you will be tempted to edit it. 
While an error make you think (most of the time). 

devil advocate, some people might stil have old versions of IPython installed ?

> if new locations already exist, do nothing

Seem legit. 

> never overwrite anything

+1

> don't copy empty config files

+1 , I suppose you consider no whitespace and comment as irrelevant. 

> small amount of find-replace in config files for renamed classes (e.g. IPythonQtConsoleApp -> JupyterQtConsoleApp)

Seem legit. 

> what files should be migrated

Profile default. Warn if other profiles ?

> copy or move, etc.? Allow overwrite?

Move (said above), no overwrite, one can nuke it's jupyter-config folder and relaunch migration if necessary. 

> API for migration (CLI)

except `jupyter migrate`, I 'm not sure what we should add. For a root user to be able to migrate each users ?

> should there be any modifications to contents, or only copies?

Minimal modif IMHO.

> When to migrate.

At Jupyter/IPython startup, if no config file, look for old IPython place, if something die and ask to run jupyter migrate. No magic involve.

minrk:
> devil advocate, some people might still have old versions of IPython installed ?

I know I will have 3.x running for some time for occasional testing. It would probably be annoying if every time I launched Jupyter, I lost all my IPython 2, 3 config.

> Move (said above), no overwrite, one can nuke it's jupyter-config folder and relaunch migration if necessary.

This actually argues for copy as well, unless I'm misunderstanding. If we move files, the originals don't exist, so nuking the new config dir means destroying the only copy, in which case relaunch will have nothing to migrate since the originals are gone.

> Profile default. Warn if other profiles ?

Agreed on profile default. Warning seems a bit much, but probably something in the output indicating that other profiles were not migrated.

> except jupyter migrate, I 'm not sure what we should add.

I'm mainly wondering if we want to have something granular, like allowing users to migrate one piece explicitly, e.g. `jupyter migrate --config` or `jupyter migrate --kernels`, etc. I'm expecting that's overkill, but I wanted to get feedback.

> At Jupyter/IPython startup

So run it at the outer-most `jupyter` command, before dispatching to subcommands? That makes sense to me. Do we want to write a `.migrated` marker file, to indicate that migration has already run, to avoid walking through a bunch of files at every invocation?

Carreau:
> So run it at the outer-most jupyter command, before dispatching to subcommands? That makes sense to me. Do we want to write a .migrated marker file, to indicate that migration has already run, to avoid walking through a bunch of files at every invocation?

Technically you try to migrate only if you don't find your own config, no ? So there is no overhead.

minrk:
Okay, that's what I misunderstood. And that would mean migration definitely is *not* handled by the outer application, but by every individual application.

It would mean that migration runs every time if the user has no config, though, which is the majority case, I think.

Carreau:
> It would mean that migration runs every time if the user has no config, though, which is the majority case, I think.

Ah, true in Jupyter config is empty by default I forgot that.

jupyter_core/migrate.py
```diff
@@ -0,0 +1,245 @@
...
+import shutil
+
+from traitlets.config import PyFileConfigLoader, JSONFileConfigLoader
+
+from ipython_genutils.path import ensure_dir_exists
+try:
+    from IPython.paths import get_ipython_dir, locate_profile
+except ImportError:
+    # IPython < 4
+    from IPython.utils.path import get_ipython_dir, locate_profile
```
Carreau: Will we ever by IPython <4 ? If so should we just abort  ?

Carreau: Or at least warn?

jupyter_core/migrate.py
```diff
@@ -0,0 +1,245 @@
...
+    dst = pjoin('{jupyter_data}', security_file)
+    migrations[src] = dst
+
+config_migrations = ['notebook', 'nbconvert', 'qtconsole']
+
+regex = re.compile
+
+config_substitutions = {
+    regex(r'\bIPythonQtConsoleApp\b'): 'JupyterQtConsoleApp',
+    regex(r'\bIPython\.html\b'): 'jupyter_notebook',
```
Carreau: `notebook`

jupyter_core/migrate.py
```diff
@@ -0,0 +1,245 @@
...
+    regex(r'\bIPython\.nbconvert\b'): 'jupyter_nbconvert',
```
Carreau: `nbconvert`

Carreau:
2 replacement since we re-renamed the repos. But +1

minrk:
@Carreau thanks, fixed the package names. Should be good to go.

PR #8, opened 2015-04-30
ensure SYSTEM_JUPYTER_PATH is never empty

minrk:
even on XP, where PROGRAMDATA is undefined, in which case fall back on sys.prefix/share/jupyter.

If there's another sensible default on XP, that would be fine, too. I'm not actually sure we support XP anymore, so this may not matter, but I don't want it possible for this list to be empty.


takluyver:
+1

PR #9, opened 2015-05-04
ensure cli config has higher priority than config files

minrk:


jupyter_core/tests/test_application.py
```diff
@@ -1,12 +1,17 @@
+import os
 from tempfile import NamedTemporaryFile
 
 from traitlets import Integer
 
 from jupyter_core.application import JupyterApp
 
+pjoin = os.path.join
```
damianavila: but `pjoin` is not used anywhere here (also the `import os` would not be necessary)... maybe a leftover of a previous implementation or I am missing something?

minrk: artefact leftover from a rebase. Removed.

takluyver: Ditto the `import os` above. And I think you forgot to push?

minrk: pushed, just to the wrong repo

jupyter_core/tests/test_application.py
```diff
@@ -27,3 +29,12 @@ def test_custom_config():
         app.initialize(['--config', f.name])
     assert app.config_file == f.name
     assert app.n == 10
+
+
+def test_cli_override():
+    app = DummyApp()
+    with NamedTemporaryFile(suffix='.py', mode='w') as f:
```
takluyver: Will this work on Windows? It needs to open two handles on the file at the same time, and I seem to remember that's problematic. We have `NamedFileInTemporaryDirectory` in genutils for this.

minrk: I'll just use mkdtemp to be safe.

PR #10, opened 2015-05-11
XDG_RUNTIME_DIR, not XDG_RUNTIME_HOME

takluyver:
See http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

Runtime is different from the other XDG directories in that it's a single directory, not a search path.


Carreau:
+1

PR #13, opened 2015-05-26
enable generating config files

minrk:
with `jupyter [app] --generate-config`

By raising `NoStart` in JupyterApp.start when start shouldn't proceed, subclasses can use `super` without having to duplicate checks for whether `start` should really do anything.

### DIFF
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -58,6 +58,8 @@ base_flags = {
             "set log level to logging.DEBUG (maximize logging output)"),
     'generate-config': ({'JupyterApp': {'generate_config': True}},
         "generate default config file"),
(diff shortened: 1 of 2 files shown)


jupyter_core/application.py
```diff
@@ -216,14 +234,22 @@ def start(self):
         """Start the whole thing"""
         if self.subcommand:
             os.execv(self.subcommand, [self.subcommand] + self.argv[1:])
-            return
+            raise NoStart()
```
takluyver: I think anything after execv is unreachable; if it succeeds, this code will no longer be executing, and if it fails, it will raise an error anyway.

takluyver:
+1

takluyver:
needs a rebase, though.

minrk:
rebased

PR #15, opened 2015-05-28
Show full subcommand names in help

### LINKED ISSUE #11: jupyter subcommands should be discoverable in help message
`ipython --help` shows a list of possible subcommands, whereas with the jupyter command it's hard to see what is available:

```console
$ jupyter --help                    
usage: jupyter [-h] subcommand

Jupyter: Interactive Computing

positional arguments:
  subcommand  The subcommand to launch

optional arguments:
  -h, --help  show this help message and exit
```

I guess it's not trivial to list the subcommands, since they are all separate programs, but it would make it easier to use.

ricklupton:
- needed to show 'install-nbextension' not 'install' in help

closes #11 ?

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -49,6 +49,7 @@ def list_subcommands():
     Returns a list of jupyter's subcommand names, without the `jupyter-` prefix.
     Nested children (e.g. jupyter-sub-subsub) are not included.
     """
+    prefix = 'jupyter-'
     path = os.environ.get('PATH') or os.defpath
     subcommands = set()
     for d in path.split(os.pathsep):
@@ -57,8 +58,8 @@ def list_subcommands():
         except OSError:
             continue
         for name in names:
-            if name.startswith('jupyter-'):
-                subcommands.add(name.split('-')[1])
+            if name.startswith(prefix):
+                subcommands.add(name[len(prefix):])
     return subcommands
 
 def main():


jupyter_core/command.py
```diff
@@ -57,8 +58,8 @@ def list_subcommands():
         except OSError:
             continue
         for name in names:
-            if name.startswith('jupyter-'):
-                subcommands.add(name.split('-')[1])
```
takluyver: `.split()` also takes an optional second parameter for the maximum number of times to split, so `name.split('-', 1)[1]` would also work correctly. I don't know that this is any neater, though.

jupyter_core/command.py
```diff
@@ -49,6 +49,7 @@ def list_subcommands():
     Returns a list of jupyter's subcommand names, without the `jupyter-` prefix.
     Nested children (e.g. jupyter-sub-subsub) are not included.
```
takluyver: This docstring seems to suggest that commands with further hyphens are treated as a hierarchical set of subcommands, so `jupyter-install-nbextension` would be invoked as `jupyter install nbextension`. But that means we can't have any subcommands with hyphens in the name. To my mind, `install-nbextension` is one thing, not a subcommand of install; if we're doing an extra level like that, I'd make it `jupyter nbextension install` (similar to kernelspecs).

@minrk ?

jupyter_core/command.py
minrk: This is deliberate, so that it only goes one level down the hierarchy. It shouldn't list every subcommand of every subcommand as a subcommand of the highest level.

jupyter_core/command.py
minrk: Yup, I would view it as a hierarchy of subcommands.

takluyver: Then should we rearrange `install-nbextension`? Or do we actually want to have an `install` subcommand namespace?

minrk:
If we want to allow subcommands with a hyphen, they should be excluded if their parent is defined. So if there's:

```
jupyter-a
jupyter-a-b
jupyter-c-d
```

It should find the subcomands `a` and `c-d`, but not `a-b`, where `b` is a subcommand of `a`.

jupyter_core/command.py
ricklupton: Rearranging `install-nbextension` into an `nbextension` subcommand seems neatest (getting rid of this PR)

takluyver:
From the dev meeting today: we'd like to leave the subcommand mechanism as is, and restructure this command as `jupyter nbextension install`.

minrk:
#29 implements and tests what I described above - if a subcommand's parent is undefined, it is added to the list with its hyphen.

PR #20, opened 2015-06-06
Simplify logic for loading config files

takluyver:
This makes it a bit easier to follow, in my opinion, and if a config file is specified with a full path, it should load it only once, whereas it was previously loaded again for each config directory.


minrk:
+5 for simplifying. Only one change: config_file should be loaded last, so it's highest priority.

minrk:
Duh, I misread and thought they were both being loaded.

PR #21, opened 2015-06-18
Pass kwargs through launch_instance()

takluyver:
nbopen uses these. The underlying Application class already has them, they just weren't exposed by this subclass.


minrk:
:+1:

PR #22, opened 2015-06-18
Be more selective with 'not a Jupyter command' error

takluyver:
This error message should only show up when the error is because the command doesn't exist. I was getting the same message for a permissions error.

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -9,6 +9,7 @@ This does nothing other than dispatch to subcommands or output path info.
 from __future__ import print_function
 
 import argparse
+import errno
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


jupyter_core/command.py
```diff
@@ -116,8 +117,10 @@ def main():
     command = 'jupyter-' + subcommand
     try:
         os.execvp(command, sys.argv[1:])
-    except OSError:
-        sys.exit("jupyter: %r is not a Jupyter command" % subcommand)
+    except OSError as e:
+        if e.errno == errno.ENOENT:
+            sys.exit("jupyter: %r is not a Jupyter command" % subcommand)
+        raise
```
minrk: I might be inclined to exit with str(e) instead of raising, to avoid the traceback. Would that be satisfactory for the permission case?

takluyver: It would be better than giving the wrong error message, certainly. But I'm inclined to leave the traceback in: errors other than ENOENT are not something we anticipate, so if one comes up, the user should have all the info to debug it. If we were Python 3 only, I'd just catch `FileNotFoundError` and let everything else bubble up.

minrk: Okay. I think the error message but no traceback would be better, but this is fine.

PR #25, opened 2015-06-23
migrate config at start of applications

minrk:
- use logging for migrate output instead of print
- use .jupyter/migrated to indicate that migration has happened,
  and shouldn't be rechecked
- default log-level is INFO, since we seem to always want that

### DIFF
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -152,12 +152,13 @@ class JupyterApp(Application):
             # already migrated
             return
 
-        ipdir = os.environ.get('IPYTHONDIR') or os.path.expanduser('~/.ipython')
+        from .migrate import get_ipython_dir, migrate
+        
+        ipdir = get_ipython_dir()
         # No IPython dir, nothing to migrate
         if not os.path.exists(ipdir):
             return
 
-        from .migrate import migrate
         migrate()
 
     def load_config_file(self, suppress_errors=True):
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -59,6 +59,9 @@ class JupyterApp(Application):
     aliases = base_aliases
     flags = base_flags
     
+    def _log_level_default(self):
+        return logging.INFO
+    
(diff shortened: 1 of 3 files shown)


jupyter_core/application.py
```diff
@@ -143,6 +146,20 @@ def ask():
         with open(config_file, mode='w') as f:
             f.write(config_text)
     
+    def migrate_config(self):
+        """Migrate config/data from IPython 3"""
+        if os.path.exists(os.path.join(self.config_dir, 'migrated')):
+            # already migrated
+            return
+
+        ipdir = os.environ.get('IPYTHONDIR') or os.path.expanduser('~/.ipython')
```
takluyver: We're importing/defining `get_ipython_dir` in the migrate module, should this import and use that rather than copying only the fallback logic?

jupyter_core/migrate.py
```diff
@@ -198,24 +209,35 @@ def migrate():
...
     
     if os.path.exists(custom_src):
-        migrate_static_custom(custom_src, custom_dst)
+        if migrate_static_custom(custom_src, custom_dst):
+            migrated = True
+    
+    # write a marker to avoid re-running migration checks
+    ensure_dir_exists(env['jupyter_config'])
+    with open(os.path.join(env['jupyter_config'], 'migrated'), 'w') as f:
+        f.write(datetime.now().isoformat())
```
takluyver: It doesn't look like we're using this timestamp anywhere, but in case we later decide to, should it be `utcnow()`? If the user changes time zone between creating and using the timestamp, calculations based on local time will be off.

jupyter_core/application.py
minrk: makes sense. Done.

jupyter_core/migrate.py
minrk: utc is always a good idea. Done.

takluyver:
Thanks, this looks good to me now.

PR #26, opened 2015-06-24
Create some docs

takluyver:
This is quite minimal, because I think most of the information about the Jupyter command, environment variables, etc. should be in the overview docs, not here. But it seems like it should have
some API docs.

Should we document any of the constants in the paths module, e.g. SYSTEM_JUPYTER_PATH, as public API? Likewise, do we consider the API of the JupyterApp base class public?


minrk:
I *don't* consider JupyterApp to be a public API, but it is an API we consume privately across Jupyter projects. I'm not sure about the constants. They can be relevant to packagers and test-environment mockers, but not really to users, who should only ever use the functions.

takluyver:
That was roughly my thinking too. I'd prefer to err on the side of not publicly documenting things when we're in doubt.

takluyver:
http://jupyter-core.readthedocs.org/en/latest/index.html

minrk:
woo!

PR #28, opened 2015-06-24
Add missing packages to setup.py

matthew-brett:


takluyver:
Gah, Python packaging. :angry: 

Thank-you for finding and fixing this.

PR #29, opened 2015-06-28
list subcommands with a hyphen in the name

minrk:
i.e. include `nbextension-install` IFF `nbextension` is not defined.

closes #15

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -76,12 +76,9 @@ def list_subcommands():
                 subcommand_tuples.add(tuple(name.split('-')[1:]))
     # build a set of subcommand strings, excluding subcommands whose parents are defined
     subcommands = set()
+    # Only include `jupyter-foo-bar` if `jupyter-foo` is not already present
     for sub_tup in subcommand_tuples:
-        for i in range(1, len(sub_tup)):
-            # Only include `jupyter-foo-bar` if `jupyter-foo` is not already present
-            if sub_tup[:i] in subcommand_tuples:
-                break
-        else:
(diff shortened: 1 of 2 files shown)


jupyter_core/tests/test_command.py
```diff
@@ -69,3 +76,25 @@ def test_subcommand_not_found():
...
+def test_subcommand_list(tmpdir):
+    a = tmpdir.mkdir("a")
+    for cmd in ('jupyter-foo-bar',
+                'jupyter-xyz',
+                'jupyter-babel-fish'):
+        a.join(cmd).write('')
+    b = tmpdir.mkdir("b")
+    for cmd in ('jupyter-foo',
+                'jupyterstuff',
+                'jupyter-yo-eyropa-ganymyde-callysto'):
```
rgbkrk: Ha

jupyter_core/command.py
```diff
@@ -72,8 +73,17 @@ def list_subcommands():
...
-    return subcommands
+                subcommand_tuples.add(tuple(name.split('-')[1:]))
+    # build a set of subcommand strings, excluding subcommands whose parents are defined
+    subcommands = set()
+    for sub_tup in subcommand_tuples:
+        for i in range(1, len(sub_tup)):
+            # Only include `jupyter-foo-bar` if `jupyter-foo` is not already present
+            if sub_tup[:i] in subcommand_tuples:
+                break
+        else:
```
takluyver: I think this check might be clearer with an any():

`if not any(sub_tup[:i] in subcommand_tuples for i in range(1, len(sub_tup))):`

minrk: Better, yes. Done.

PR #30, opened 2015-07-01
Copy cl_config before loading config from file

takluyver:
Avoids problem with traitlets master: without this, cl_config is the same as `self.config`, so it's changed by loading the config file, and doesn't store the config values from the command line.


minrk:
:+1: merge when Travis agrees.

PR #41, opened 2015-08-13
Use setuptools entry points when setuptools in use

takluyver:
Part of the fix for jupyter/jupyter#15 - using plain 'scripts', we don't
get the exe wrappers on Windows.

To fix it completely, we can either make the same change in all packages
installing scripts (notebook, nbconvert, qtconsole, console), or make
the recognition of subcommands a bit smarter to know about running
Python scripts without an exe wrapper. I'm inclined to do the former -
more pain right now, but probably less in the long term.

 @minrk, thoughts

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,9 +68,17 @@ setuptools_args['install_requires'] = [
     'traitlets',
 ]
 
+setuptools_args['entry_points'] = {
+    'console_scripts': [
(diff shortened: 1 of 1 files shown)


minrk:
Windows + Setuptools = sad Min. I'd like to see what making it work without using entrypoints everywhere would look like before committing to it. But I don't think we can avoid it on this repo, at least.

Issue #47, opened 2015-08-26
zsh autocompletion file


anntzer:
I have written a small zsh autocompletion file for jupyter, currently at http://pastebin.com/XW0nPWjG .
There seemed to be some interest in such a file for ipython (cf ipython/ipython#8767), but I guess it would make more sense to merge this into jupyter_core?
I haven't made this into a PR because I don't know where to put it in the source tree :-)

Carreau:
Well if it works with jupyter, +1, though the completion will work even if subcommand not availlable ?

I would create a `/tools` or whatever folder with this in it. 

conda have a nice `conda install argcomplete` do we want something similar ?

anntzer:
Good catch, I just wrote this for whatever commands got installed by `<linux-package-manager> install jupyter`.  A bit {la,bu}zy to test installing subcommands one at a time right now (I guess I can just `pip install jupyter_core && pip install <subcommand>` in a clean conda env?) but it should boil down to a bunch of `if which jupyter-$subcmd; then commands=($commands $subcmd_help); fi` in `_jupyter_cmds`.

Carreau:
> A bit {la,bu}zy to test installing subcommands one at a time right now

Well I think it is fine if it completes and the subcommand is not availlable. It's not the end of the world, 
and you are likely to know which command are availlable anyway.

anntzer:
fixed it.

Carreau:
Closed by #48

PR #48, opened 2015-08-26
A zsh autocompleter for jupyter.

anntzer:
See #47.

### DIFF
--- a/tools/_jupyter
+++ b/tools/_jupyter
@@ -1,0 +1,193 @@
+#compdef jupyter
+# A zsh autocompleter for jupyter.
+_jupyter() {
+  IFS=$'\n'
+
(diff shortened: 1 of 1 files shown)


Carreau:
Note to check wether setup.py include that in the sdist (and wheel?) before merging. 

+1 otherwise.

anntzer:
Doesn't seem to be included.  Not really sure whether it should be or not: either you can have the users include the folder in their `$fpath` (in `.zshrc`/`.zprofile`), or rely on the downstream packages to dump the file in (e.g., for Arch) `/usr/share/zsh/functions/Completion/...`

Carreau:
don't worry we'll modify setup.py to at least include that in sdist. I just
dropped the line as a reminder for us.

PR #54, opened 2015-09-12
don't use exec on Windows

minrk:
Use regular Popen instead.

Windows doesn't actually allow process replacement a la unix exec, but Python provides os.exec on Windows even though it doesn't actually work. It doesn't exec, it does fork & exec, which means the parent process returns immediately, and the 'true' process is put in the background.

The alternative is to use execfile on Windows, which would behave the same as exec on \*ix, with one exception: it forces all jupyter commands to be run with the same interpreter, rather than respecting shebang lines.

closes jupyter/nbconvert#103

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -13,6 +13,7 @@ import errno
 import json
 import os
 import sys
+from subprocess import Popen
 
 from . import paths
 from .version import __version__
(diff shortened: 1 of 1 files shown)


minrk:
Confirmed working on Windows.

sccolbert:
Link to proof that `POpen` handles args with spaces on Windows:
https://github.com/python/cpython/blob/master/Lib/subprocess.py#L715
https://github.com/python/cpython/blob/master/Lib/subprocess.py#L754

minrk:
Thanks, @sccolbert!

Carreau:
I guess that also fixes jupyter/nbconvert#103 ?

Carreau:
> I guess that also fixes jupyter/nbconvert#103 ?

Hah, yes already closed. The page was still opened for me and did not refresh correctly. Thanks.

jorisvandenbossche:
This also fixes jupyter console on windows: https://github.com/jupyter/jupyter_console/issues/26 (based on a quick test it seems fixed)

PR #62, opened 2015-09-25
ensure jupyter dir is on PATH

minrk:
before dispatching to subcommands

This addresses the case when jupyter is executed by absolute path and its dir is not on PATH, which would result in all subcommands not being found.

closes jupyter/notebook#448

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -56,6 +56,7 @@ def jupyter_parser():
     
     return parser
 
+
 def list_subcommands():
     """List all jupyter subcommands
     
@@ -64,10 +65,9 @@ def list_subcommands():
     Returns a list of jupyter's subcommand names, without the `jupyter-` prefix.
     Nested children (e.g. jupyter-sub-subsub) are not included.
     """
-    path = os.environ.get('PATH') or os.defpath
     subcommand_tuples = set()
     # construct a set of `('foo', 'bar') from `jupyter-foo-bar`
-    for d in path.split(os.pathsep):
+    for d in _path_with_self():
         try:
             names = os.listdir(d)
         except OSError:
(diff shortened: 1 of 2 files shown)


bbergman:
Weird that I JUST hit this today when I tried to create an init.d entry for notebook. Thankfully, the workaround works. :)

EvanZ:
Which dir?

root@5094eba8037a:/# jupyter --paths
config:
    /root/.jupyter
    /usr/etc/jupyter
    /usr/local/etc/jupyter
    /etc/jupyter
data:
    /root/.local/share/jupyter
    /usr/local/share/jupyter
    /usr/share/jupyter
runtime:
    /root/.local/share/jupyter/runtime

takluyver:
The directory where the Jupyter command itself is. See `dirname $(which jupyter)`.

ibrahim5253:
I am on a server (without sudo access) and facing this issue. But `ipython notebook` works fine. Also, `jupyter`'s dir (`\usr\local\bin`) is in the `PATH` variable. What can be done?

takluyver:
Where is the `jupyter-notebook` script?

ibrahim5253:
I doubt there exists any. `which jupyter-notebook` returns nothing.

takluyver:
If that doesn't exist, then `jupyter notebook` won't work. You may be able to install the notebook without sudo access, depending on how the server is configured.

stevengj:
This still failing for me in Windows 8: when I run `\somepath\jupyter.exe notebook` it fails with "jupyter-notebook not found`, but when I add `\somepath` to the `PATH` environment variable it works.  It also works if I run `\somepath\jupyter-notebook.exe` directly.

`jupyter --version` prints 4.4.0.

PR #64, opened 2015-10-12
import typo in test_command

minrk:
caused py2 import branch on all Python versions

This wasn't noticed because the `mock` package was installed anyway.


willingc:
LGTM

PR #65, opened 2015-10-17
use shell=True on Windows

adalq:
You need to pass `shell=True` on Windows, otherwise the command won't be
looked up in `PATH`.

 Without this fix I get the `jupyter: 'notebook' is not a Jupyter command` error when executing any jupyter command. Tested with latest version (4.0.6)

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -93,7 +93,7 @@ def _execvp(cmd, argv):
     Python provides execvp on Windows, but its behavior is problematic (Python bug#9148).
     """
     if sys.platform.startswith('win'):
-        p = Popen([cmd] + argv[1:])
+        p = Popen([cmd] + argv[1:], shell=True)
         # Don't raise KeyboardInterrupt in the parent process.
         # Set this after spawning, to avoid subprocess inheriting handler.
         import signal


minrk:
Hm, I wonder why this has been working for others on Windows. Can you share your `pip list` or `conda list`?

adalq:
For a bunch of reasons I install packages the manual way (by unzipping the `.whl` in the proper place). So I don't have `pip` or `conda` installed on my system.

I understand if you don't want to support this scenario, and I've noticed lately that a few packages don't work anymore when they are not installed through a package manager (for example because they rely on advanced `pbr` functionality).

minrk:
I'm pretty sure that unzipping a wheel is not *quite* enough to replicate the installation of entrypoints. Can you share the listing of files in your script directory (where `jupyter-notebook` would be)?

adalq:
You're right, I had to manually create the entrypoints:

    ipython.bat
    jupyter.bat
    jupyter-console.bat
    jupyter-kernelspec.bat
    jupyter-nbextension.bat
    jupyter-notebook.bat
    jupyter-qtconsole.bat
    nb.bat
    ipython.ico
    ipython_nb.ico
    QtConsole.lnk

I looked into the `entry_points.txt` file from the `.whl` to figure out what to import for each of them, and also noticed that most packages have a `__main__.py`. `%*` forwards command line arguments.

    # jupyter-kernelspec.bat:
    python -m jupyter_client.kernelspecapp %*

    # jupyter-notebook.bat:
    python -m notebook %*

    # and so on ...

BTW, this whole mechanism - `jupyter_core` package searching `PATH` for executables starting with `jupyter-` which were created by the install scripts (thus requesting extra steps for manual installations) feels fragile and also a bit circular. Especially the search the `PATH` part, which doesn't look right from a security point of view (although I can't pinpoint a specific attack, since if someone can put files in your `PATH`, you're screwed anyway). But I guess you had your reasons for doing things this way.

dmitryponv:
where do you pass shell=True, Im having an issue with Jupyter not recognizing ffmpeg
https://stackoverflow.com/questions/60874064/jupyter-not-recognizing-environment-variables-in-windows-that-are-working-in-pyt

PR #66, opened 2015-10-22
Use unicode everywhere, because unicode is good, python2 is bad

Carreau:
Pretty sure that will fix jupyter/notebook#646

Need to write a test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: python
 sudo: false
 python:
+    - 3.5
     - 3.4
     - 3.3
     - 2.7
--- a/jupyter_core/migrate.py
+++ b/jupyter_core/migrate.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 """Migrating IPython < 4.0 to Jupyter
 
 This *copies* configuration and resources to their new locations in Jupyter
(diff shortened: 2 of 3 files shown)


minrk:
I don't love using unicode_literals, since it results in some pretty surprising behavior, but if this is the most direct fix, +1. Looking forward to the test.

Carreau:
> Looking forward to the test.

I'll investigate on monday.

Carreau:
Ok, I have a failing test but that fails for the wrong thing :-P 

Will investigate deeper later.

Carreau:
What a joy. 

os.envion on 3 have unicode values and keys that use `getsystemencoding`. 

use `environb` for bytes. 

Python2 has only `os.environ` it is not sure which one, seem to be bytes !

willingc:
@Carreau For more joy and pain, http://nedbatchelder.com/text/unipain.html

Carreau:
Yeah, I've read Ned's talk and watched the talk twice I'm annoyed because the fix is long and annoying. 

I would be in favor in not supporting python2.7+unicode home dirs.

As pointed by Aaron Meurer on Twitter, anyway you can't have set unicode env variable on windows, so I'm not sure it is worth pursuing on other platform.

Carreau:
Ok, I gave up on writing for unicode home, though this still test the behavior for #646.

PR #72, opened 2016-02-20
troubleshoot script from notebook Issue #1103

rmanly:
Collects and prints out the data requested in the comments for [notebook ISSUE #1103](https://github.com/jupyter/notebook/issues/1103).

The revision history was extremely short and I didn't think it was worth migrating the repo from my fork of notebook to here after the location change was suggested.

There is one small shortcoming in that I don't know what Windows installations would do with `which -a jupyter` and I did not attempt to address that at all.

### DIFF
--- a/scripts/jupyter-troubleshoot
+++ b/scripts/jupyter-troubleshoot
@@ -23,7 +23,7 @@ def subs(cmd):
     """
     try:
         stdout = subprocess.check_output(cmd)
-        return stdout
+        return stdout.decode('utf-8')
     except subprocess.CalledProcessError:
         return None
 
--- a/scripts/jupyter-troubleshoot
+++ b/scripts/jupyter-troubleshoot
@@ -1,0 +1,90 @@
+#!/usr/bin/env python
+"""
+display environment information that isfrequently
+used to troubleshoot installations of Jupyter or IPython
+"""
+
+# import argparse
+import os
+import platform
+import subprocess
+import sys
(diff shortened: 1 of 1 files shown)


rgbkrk:
Thanks for the contribution!

takluyver:
Thanks, I think this is a good start. Do you have more time to spend on it, or would you rather we merged this and other people work on it?

Where it's available, i.e. on Python 3, we can use `shutil.which()`, which will work on Windows as well. It has no equivalent to `which -a`, though.

rmanly:
I am leaving the sprint event in a few minutes but I can work on this some more tomorrow.

But by all means merge if you feel it is worthwhile ;)

Carreau:
> I am leaving the sprint event in a few minutes but I can work on this some more tomorrow.
> 
> But by all means merge if you feel it is worthwhile ;)

Thanks ! I'm going to merge this as is, so that we (and you) can iterate on this !

rmanly:
Thanks.

Some major life events got in the way over the last few days. So I won't have much time to iterate soon. It will be nice to have some folks who have dealt with cross-platform and version issues take a look.

Carreau:
> Some major life events got in the way over the last few days. So I won't have much time to iterate soon. It will be nice to have some folks who have dealt with cross-platform and version issues take a look.

No problem, take your time. We'll work  on that in time.

Thanks to you !

PR #74, opened 2016-03-07
Prepare for 4.1

minrk:
- Added changelog notes
- Some fixes for the troubleshoot script
  - handle missing data
  - ensure script is installed


docs/changelog.rst
```diff
@@ -1,6 +1,21 @@
...
+4.1
+---
+
+4.1.0
+~~~~~
+
+`on
+GitHub <https://github.com/jupyter/jupyter_core/releases/tag/4.1>`__
+
+- Add `jupyter.py` module, so that :command:`python -m jupyter` always works.
```
takluyver: This repo doesn't have the single-backticks-is-double-backticks thing enabled. If you want it, the setting is here: https://github.com/jupyter/jupyter_core/blob/7d71e36d62a52e2c55593a092177447ff53b3995/docs/conf.py#L87

takluyver: (IIRC, set it to `'literal'`)

jupyter_core/troubleshoot.py
```diff
@@ -24,7 +24,7 @@ def subs(cmd):
     try:
         stdout = subprocess.check_output(cmd)
         return stdout.decode('utf-8').strip()
```
takluyver: While we're making this more robust, let's use the 'replace' error handler for decoding, so it won't crash if it gets non-UTF-8 output.

jupyter_core/troubleshoot.py
```diff
@@ -74,16 +74,20 @@ def main():
     print('\n' + 'platform.platform():')
     print('\t' + environment_data['platform'])
 
-    print('\n' + 'which -a jupyter:')
-    print('\t' + environment_data['which'])
+    if environment_data['which']:
+        print('\n' + 'which -a jupyter:')
```
takluyver: There is a `where` command we could use on Windows, or `shutil.which()`. I don't think either of them have an equivalent of the `-a` flag, though.

docs/changelog.rst
minrk: Fixed

jupyter_core/troubleshoot.py
minrk: Done

jupyter_core/troubleshoot.py
minrk: Looks like `where` is already `which -a`, so it's exactly what we want. Yay, Windows!

minrk:
Thanks! I'll make the 4.1 release.

minrk:
Released.

willingc:
@minrk @takluyver Nice job on the release :stars: I do have a question related to what you would like as the default version on RTD. It would be good to set RTD to build this tag and make it the stable version as well in RTD. Also, the latest version is the current default for display. Perhaps it would be better for stable to be the current default.

I'm happy to do the RTD changes, if added as a doc admin.

takluyver:
I personally find more annoyances when the default docs view is 'stable' and improvements to the docs are hidden, than when it's 'latest' and describes some things that are only in the development version. Of course, the less improvement our docs need, the better the argument for 'stable' being the default ;-)

minrk:
I *tend* to prefer 'latest' as the default docs, but that may not be the right choice for users. I think if we're good about marking things with `versionchanged`, then latest docs will typically be better than stable docs, since not-yet-released things should be clearly marked, and other improvements that aren't version-sensitive will not be waiting for a code release. But I don't know if having unreleased APIs would be confusing to people.

@willingc I did make you a doc admin, so you should be able to make any changes, whatever the right ones turn out to be.

willingc:
I think in this case 'latest' will be fine. If we make extensive changes going forward, we can revisit this.

Thanks @minrk @takluyver

PR #77, opened 2016-05-03
Properly set PATH when `jupyter` is called from a symlink

scolby33:
When calling `jupyter` that was installed inside a virtual environment while that environment is not active, it is unable to locate the other executables, like `jupyter-notebook` because `execvp` uses `$PATH` instead of `sys.path` to locate the executable. `sys.path` is automatically updated by the Python interpreter to add the path to the virtual environment binaries and libraries and `$PATH` is not.

This change prepends the `$PATH` environment variable with the first entry from `sys.path` before calling `execvp` to rectify this.

NOTE: I am unfamiliar with Windows' handling of environment variables and its path, so the Windows case should be updated as well before this is merged.

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -106,17 +106,20 @@ def _execvp(cmd, argv):
 
 def _path_with_self():
     """Ensure `jupyter`'s dir is on PATH"""
-    script = sys.argv[0]
-    bindir = os.path.dirname(script)
+    scripts = [sys.argv[0]]
+    if os.path.islink(scripts[0]):
+        scripts.append(os.path.realpath(scripts[0]))
+    bindirs = [os.path.dirname(script) for script in scripts]
     path_list = (os.environ.get('PATH') or os.defpath).split(os.pathsep)
-    if (os.path.isdir(bindir)
-        and bindir not in path_list
-        and os.access(script, os.X_OK) # only if it's a script
-    ):
-        # ensure executable's dir is on PATH
-        # avoids missing subcommands when jupyter is run via absolute path
-        path_list.append(bindir)
-        os.environ['PATH'] = os.pathsep.join(path_list)
+    for bindir, script in zip(bindirs, scripts):
+        if (os.path.isdir(bindir)
+            and bindir not in path_list
+            and os.access(script, os.X_OK) # only if it's a script
+        ):
+            # ensure executable's dir is on PATH
+            # avoids missing subcommands when jupyter is run via absolute path
+            path_list.append(bindir)
+            os.environ['PATH'] = os.pathsep.join(path_list)
     return path_list
 
 
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -109,9 +109,9 @@ def _path_with_self():
     scripts = [sys.argv[0]]
     if os.path.islink(scripts[0]):
         scripts.append(os.path.realpath(scripts[0]))
-    bindirs = [os.path.dirname(script) for script in scripts]
     path_list = (os.environ.get('PATH') or os.defpath).split(os.pathsep)
-    for bindir, script in zip(bindirs, scripts):
+    for script in scripts:
+        bindir = os.path.dirname(script)
         if (os.path.isdir(bindir)
             and bindir not in path_list
             and os.access(script, os.X_OK) # only if it's a script
@@ -119,7 +119,7 @@ def _path_with_self():
             # ensure executable's dir is on PATH
             # avoids missing subcommands when jupyter is run via absolute path
             path_list.append(bindir)
-            os.environ['PATH'] = os.pathsep.join(path_list)
+    os.environ['PATH'] = os.pathsep.join(path_list)
     return path_list
 
 


takluyver:
I think what you want should already have been implemented in #62 - check that you have the latest version of `jupyter_core` installed. I don't think adding the first thing on `sys.path` is the right way to fix this.

scolby33:
I apologize for being hasty with my pull request--I am indeed on the latest version and saw that code, but didn't take the time to think about what it did. I agree that this is not the best way to fix my issue. I'll describe my current setup to make the issue clear.

```bash
$ echo $PATH
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/MacGPG2/bin:/Library/TeX/texbin:/opt/coin/bin:/Users/scott/.local/bin

$ which jupyter
~/.local/bin/jupyter

$ tree -aI .DS_Store
~
└── .local
    ├── bin
    │   └── jupyter -> ~/.local/venvs/jupyter/bin/jupyter
    └── venvs
        └── jupyter
            ├── bin
            │   ├── jupyter
            │   ├── jupyter-console
            │   ├── jupyter-etc
            │   ├── jupyter-notebook
            │   └── python3.5 -> /usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/bin/python3.5
            └── lib
                └── python3.5
                    └── site-packages
                        ├── jupyter.py
                        ├── jupyter_core
                        │   ├── command.py
                        │   └── etc.py
                        └── jupyter_etc
```
I have the components of `jupyter` installed in a virtualenv under `~/.local/venvs` and have symlinked `jupyter` into `~/.local/bin` which is on my path. The `jupyter` file in `~/.local/venvs/jupyter/bin` looks like this:
```python
#!~.local/venvs/jupyter/bin/python3.5
# EASY-INSTALL-ENTRY-SCRIPT: 'jupyter-core==4.2.0.dev0','console_scripts','jupyter'
__requires__ = 'jupyter-core==4.2.0.dev0'
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('jupyter-core==4.2.0.dev0', 'console_scripts', 'jupyter')()
    )
```
The important part here is the hashbang line, which makes the shell run it from the Python installed in the virtualenv.

Here's what happens when I run this file with various different invocations. I have added some `print` statements inside `jupyter_core.command` to debug what's going on, showing $PATH and sys.path in their initial states and immediately before the `exec()` call as well as a notification when the script enters inside the if statement in `_path_with_self`.
```bash
$ jupyter notebook
sys.path in main:
['/Users/scott/.local/venvs/jupyter/bin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python35.zip', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/plat-darwin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/lib-dynload', '/Users/scott/.local/venvs/jupyter/lib/python3.5/site-packages']
os.environ['PATH'] in main:
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/MacGPG2/bin:/Library/TeX/texbin:/opt/coin/bin:/Users/scott/.local/bin
sys.path in _execvp:
['/Users/scott/.local/venvs/jupyter/bin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python35.zip', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/plat-darwin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/lib-dynload', '/Users/scott/.local/venvs/jupyter/lib/python3.5/site-packages']
os.environ['PATH'] in _execvp:
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/MacGPG2/bin:/Library/TeX/texbin:/opt/coin/bin:/Users/scott/.local/bin
jupyter: 'notebook' is not a Jupyter command

$ ~/.local/bin/jupyter notebook
sys.path in main:
['/Users/scott/.local/venvs/jupyter/bin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python35.zip', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/plat-darwin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/lib-dynload', '/Users/scott/.local/venvs/jupyter/lib/python3.5/site-packages']
os.environ['PATH'] in main:
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/MacGPG2/bin:/Library/TeX/texbin:/opt/coin/bin:/Users/scott/.local/bin
sys.path in _execvp:
['/Users/scott/.local/venvs/jupyter/bin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python35.zip', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/plat-darwin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/lib-dynload', '/Users/scott/.local/venvs/jupyter/lib/python3.5/site-packages']
os.environ['PATH'] in _execvp:
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/MacGPG2/bin:/Library/TeX/texbin:/opt/coin/bin:/Users/scott/.local/bin
jupyter: 'notebook' is not a Jupyter command

$ ~/.local/venvs/jupyter/bin/jupyter notebook
sys.path in main:
['/Users/scott/.local/venvs/jupyter/bin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python35.zip', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/plat-darwin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/lib-dynload', '/Users/scott/.local/venvs/jupyter/lib/python3.5/site-packages']
os.environ['PATH'] in main:
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/MacGPG2/bin:/Library/TeX/texbin:/opt/coin/bin:/Users/scott/.local/bin
Inside _path_with_self if statement
sys.path in _execvp:
['/Users/scott/.local/venvs/jupyter/bin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python35.zip', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/plat-darwin', '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/lib-dynload', '/Users/scott/.local/venvs/jupyter/lib/python3.5/site-packages']
os.environ['PATH'] in _execvp:
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/MacGPG2/bin:/Library/TeX/texbin:/opt/coin/bin:/Users/scott/.local/bin:/Users/scott/.local/venvs/jupyter/bin
/Users/scott/.local/venvs/jupyter/lib/python3.5/site-packages/widgetsnbextension/__init__.py:30: UserWarning: To use the jupyter-js-widgets nbextension, you'll need to update
    the Jupyter notebook to version 4.2 or later.
  the Jupyter notebook to version 4.2 or later.""")
[I 09:46:55.336 NotebookApp] Serving notebooks from local directory: /Users/scott
... etc. everything works from here
```
In the first two versions of the invocation, the actual location of the `jupyter` binaries/subcommands is never on the path because of the symlink. In fact, the if statement in `_path_with_self` doesn't evaluate to True in those first two, but does go to True in the last version.

Also note that the Python interpreter automatically adds several things to `sys.path` that weren't present in `$PATH`.

So, in the end `_path_with_self` doesn't work quite right if `jupyter` is called via a symlink. Perhaps the solution is to check if `sys.argv[0]` is a symlink and follow it before adding to the path?

I have attached my version of `command.py` with the extra print statements (as a text file because Github didn't like the .py extension): [command.py.txt](https://github.com/jupyter/jupyter_core/files/249217/command.py.txt)

Apologies for the wall of text and command output, but I hope this was helpful in defining my problem.

scolby33:
After taking the time to write out that big comment above, a (reasonable, I think) solution became clear to me. Here's a one-line change that implements that and I think fits into the framework of the code much better than my first attempt.

takluyver:
Hmmm - you could also have `jupyter` and the `jupyter-*` scripts symlinked together into one place from different places, in which case this would break something that currently works. I can't think of an obvious use case for that, though.

takluyver:
(Note for merging this, we should squash to cancel out the redundant change)

scolby33:
That's a good point. Perhaps instead of just adding the `realpath`, both `sys.argv[0]` and `realpath` should be added. This would cause an extra search on the path in one or the other case, but that's not the end of the world.

I would put `sys.argv[0]` on first, followed by the `realpath` only if `sys.argv[0]` is a symlink. That way if anyone currently has the `jupyter-*` scripts symlinked in from different places they won't notice a change, but if they can't be found there, it'll proceed to the actual location of the `jupyter` script.

takluyver:
Pinging @minrk for thoughts on this. I think searching the link location first and then the realpath location probably makes sense.

minrk:
Symlink then real path makes some sense, but I'm confused as to how you get into the situation where this makes a difference. It seems like that shouldn't happen unless something is going wrong.

takluyver:
@minrk I think the key is that `jupyter` itself is a symlink, not the directory containing it. And the other jupyter-* scripts are not symlinked. I do something like this with conda - I don't add its bin to PATH, but I symlink `conda` and `activate` into a location that is on PATH.

scolby33:
Here's a first pass at putting both `sys.argv[0]` and its `realpath` on the path. I'm not 100% crazy about the implementation, but I've put it up for comment.

minrk:
Thanks, @scolby33! Looks okay to me.

scolby33:
Squashed commits

jupyter_core/command.py
```diff
@@ -106,17 +106,20 @@ def _execvp(cmd, argv):
...
     path_list = (os.environ.get('PATH') or os.defpath).split(os.pathsep)
-    if (os.path.isdir(bindir)
-        and bindir not in path_list
-        and os.access(script, os.X_OK) # only if it's a script
-    ):
-        # ensure executable's dir is on PATH
-        # avoids missing subcommands when jupyter is run via absolute path
-        path_list.append(bindir)
-        os.environ['PATH'] = os.pathsep.join(path_list)
+    for bindir, script in zip(bindirs, scripts):
```
takluyver: I found this a little tricky to follow. Rather than computing bindirs and then zipping it back together with scripts, I think it would be clearer if we iterated over it once:

```python
for script in scripts:
    bindir = os.path.dirname(script)
    if (os.path.isdir(bindir)
       ...
```

scolby33:
Agreed. See my newest change.

takluyver:
Thanks @scolby33 :-)

minrk:
Thanks!

PR #79, opened 2016-06-30
[proposal] Prioritize jupyter script directory for subcommands

minrk:
ensure that `/path/to/jupyter-subcommand` has top priority when `/path/to/jupyter subcommand` is called.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ __pycache__
 .#*
 .coverage
 htmlcov
+.cache
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -105,20 +105,26 @@ def _execvp(cmd, argv):
 
 
 def _path_with_self():
-    """Ensure `jupyter`'s dir is on PATH"""
+    """Put `jupyter`'s dir at the front of PATH
+    
+    Ensures that /path/to/jupyter subcommand
+    will do /path/to/jupyter-subcommand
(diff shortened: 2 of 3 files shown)


minrk:
We have to choose between:

- `jupyter subcommand` is always equivalent to `jupyter-subcommand` (trust PATH), or
- `/path/to/jupyter subcommand` is always equivalent to `/path/to/jupyter-subcommand` (if present, overrides PATH)

upside: explicit `/path/to/jupyter subcommand` behaves how people seem to expect, in terms of dispatch when another Jupyter installation may be ahead on PATH.

(theoretical) downside: writing your own `jupyter-subcommand` and putting it at a high priority on PATH no longer overrides a subcommand installed in the same directory as `jupyter` itself. This could come up if there is a system-wide installation of a `jupyter-whatever` tool, but the user wants her own installation of that particular tool to take precedence. This would only be possible if she also has the base `jupyter` executable in her directory (likely if it's a Python package that depends on jupyter_core, not necessarily otherwise).

minrk:
@takluyver want to look at this one? I'll cut 4.2 with #88 shortly, but this one has been sitting around a long time, so I can include it, too, if you think it makes sense.

takluyver:
I think the way this changes to makes more sense. Overriding subcommands would still be possible, just a bit more effort. So let's go for it.

minrk:
Roughly what I was thinking, too, thanks!

PR #80, opened 2016-06-30
Bring bash completion to jupyter form IPython.

Carreau:
As requested from @fperez


fperez:
Thanks! This looks inocuous enough and at least we move it to the right place, even if it can be improved later...

ps - title of PR needs fixing: 'bach' -> 'bash'.

fperez:
Never mind, I forgot I could fix the title myself, done.

takluyver:
I think we could sell a Bach completion tool to composers ;-)

PR #95, opened 2017-01-03
More informative error message if _execvp fails

jdemeyer:
It happened more than once that the error message of a failing `_execvp` was unclear to me. This branch helps that.


minrk:
Thanks!

PR #96, opened 2017-01-23
Add environment variable to ignore all config

takluyver:
Setting JUPYTER_NO_CONFIG=1 will ignore all custom config. I'm using the convention of taking any non-empty value as True.

Closes jupyter/notebook#1992

I also considered that maybe this should ignore anything in the data search path as well (in which case, the environment variable may need a different name). However, while nbextensions are installed in data
files, they are only loaded if they are enabled in config, so I think this should avoid loading them. We may still want to make it affect the data path too for future-proofing.

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -24,6 +24,18 @@ def get_home_dir():
     homedir = os.path.realpath(homedir)
     return homedir
 
+_dtemps = {}
+def _mkdtemp_once(name):
+    """Make or reuse a temporary directory.
+
+    If this is called with the same name in the same process, it will return
+    the same directory.
+    """
+    try:
+        return _dtemps[name]
+    except KeyError:
+        d = _dtemps[name] = tempfile.mkdtemp(prefix=name + '-')
+        return d
 
 def jupyter_config_dir():
     """Get the Jupyter config directory for this platform and user.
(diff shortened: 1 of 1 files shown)


jupyter_core/paths.py
```diff
@@ -33,6 +34,9 @@ def jupyter_config_dir():
     env = os.environ
     home_dir = get_home_dir()
 
+    if env.get('JUPYTER_NO_CONFIG'):
+        return tempfile.mkdtemp(prefix='jupyter-clean-cfg-')
```
minrk: If this isn't cached, every call to `jupyter_config_dir` will create a new directory and not clean it up. Maybe store it somewhere?

minrk:
This definitely covers most cases, and I think it makes sense in general to separate ignoring config from ignoring data.

In the future, it may still make sense to enable excluding data as well, though that is significantly less important. I've had plenty of kernelspec-related tests fail due to what I have installed, for instance. In general, we may want to be able to force the paths to be an explicit value, excluding all defaults, e.g. something like `JUPYTER_ENV_STRICT` that only allows fields on the `JUPYTER_*_PATH` variables, ignoring user and system defaults.

Switches people seem to want to flip:

- no user files (config and/or data)
- no system files (config and/or data)
- sys-prefix-only (strict virualenvs)

jupyter_core/paths.py
takluyver: Good point, done.

Carreau:
+1

maya-harel:
where should this env variable set ? on the machine that the jupyterlab is running on ?

PR #98, opened 2017-02-15
[zsh-completion] copy&paste mistake in description

pseyfert:
when tab completing "jupyter notebook <TAB>" the completion suggestions should be titled "notebook command" rather than "nbextension command"


takluyver:
Thanks

Issue #99, opened 2017-03-01
Dependency on ipython is circular and not necessary


tomdz:
`jupyter_core` currently depends on `ipython` in dev mode. However this introduces a circular dependency since ipython depends on `jupyter_core` for tests (via `ipykernel` -> `jupyter_client` -> `jupyter_core`).
Looking at the code, this dependency actually seems completely unnecessary, the only place where it is used, is here: https://github.com/jupyter/jupyter_core/blob/master/jupyter_core/tests/test_migrate.py#L19-L22
and `locate_profile` is not used anywhere.

rgbkrk:
The only part where I think it might be necessary is for having the default kernel available in tests (which would mean a requirement on `ipykernel` though...)

rgbkrk:
Also, thanks for raising this issue! Good to have some cleanup here.

PR #100, opened 2017-03-01
locate_profile is not used in this test

rgbkrk:
Hat tip to @tomdz as part of #99.


ivanov:
Thanks @rgbkrk and @tomdz! 🍰 for you both!

Issue #111, opened 2017-08-26
Added license file


tdalseide:

ivanov:
A version of this landed in #344, closing.

PR #112, opened 2017-08-26
Added license file.

tdalseide:
#111


Carreau:
Thanks for the PR ! We should likely move Copying.md to LICENSE instead of creating it from scratch and in particular keep the copyright year (2015-Present) instead of moving 2017.

PR #113, opened 2017-10-11
Add utility function ensure_dir_exists

takluyver:
See discussion on https://github.com/jupyter/notebook/pull/2924

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,6 @@ install:
   - pip install -r dev-requirements.txt .
 script:
   - py.test jupyter_core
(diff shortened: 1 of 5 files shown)


takluyver:
The test failure is an issue with compiling zmq on Python nightly. I'm not sure that it's really worth testing this project on Python nightly.

minrk:
👍 this repo does use ensure_dir_exists several places, so it probably makes sense to update those as well.

takluyver:
Updated, and also allowed builds on Python nightly to fail.

PR #115, opened 2017-10-17
Use standard locations for environment path

blink1073:
When using `data_files` in `setup.py`, files are stored in `/usr/local`.  

cf https://github.com/jupyterlab/jupyterlab/issues/3105

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -174,7 +174,10 @@ else:
         "/etc/jupyter",
     ]
 
-ENV_CONFIG_PATH = [os.path.join(sys.prefix, 'etc', 'jupyter')]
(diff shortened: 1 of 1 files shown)


blink1073:
Ha, it turns out that this isn't standard at all.   We had a user report that the data files ended up in `/usr/share` on their system.

https://github.com/jupyterlab/jupyterlab/issues/3123#issuecomment-337950678

takluyver:
It's complicated. On Linux distros, typically `/usr` contains files installed through the package manager, and `/usr/local` has files installed outside that. There's also some cases on Mac where `sys.prefix` can be 'wrong' for looking up data files, but I forget what cases those are.

This is part of the reason why we use a search path which includes both a directory relative to `sys.prefix` and hardcoded system locations.

PR #118, opened 2017-10-26
Avoid triggering extra logic in IPython.paths.get_ipython_dir when performing migration

minrk:
IPython's get_ipython_dir always returns a writable directory that exists, creating one if the default resolution isn't writable.

We don't want to trigger all of this extra logic when checking for old config files to migrate,
so reimplement a simpler IPython dir resolution without any of IPython's own migration logic.

We already had a check if `get_ipython_dir()` exists to skip migration, but this would always be True because of upstream `get_ipython_dir()`'s behavior.

closes jupyterhub/jupyterhub#1503

It's also been long enough that we could consider removing automatic migration altogether, leaving only the explicit `jupyter-migrate` command.

### DIFF
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -160,10 +160,9 @@ class JupyterApp(Application):
             return
 
         from .migrate import get_ipython_dir, migrate
-        
-        ipdir = get_ipython_dir()
+
         # No IPython dir, nothing to migrate
-        if not os.path.exists(ipdir):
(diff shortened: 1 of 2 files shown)


takluyver:
Makes sense. I'd also be OK with dropping auto-migration, since Jupyter has been around over two years. But I don't think it does much harm to leave it around either.

Carreau:
Happy to drop auto-migration at some point in the future, it is still cheap to keep.

PR #119, opened 2017-10-31
Changelog for 4.4.0

takluyver:
The last release was in February, and there have been a few changes in master, so I'm planning to do a new release.


Carreau:
> The last release was in February, and there have been a few changes in master, so I'm planning to do a new release.

Thanks !

PR #125, opened 2018-02-07
fix bash-completion code

andram:
When transferring from ipython, the jupyter-completion.bash file was
not adapted correctly, leading to a segfault. This patch fixes the
issue.


Carreau:
Oh my... segfault !?

Thanks.

PR #126, opened 2018-02-08
bash completion tweaks

andram:
Bash completion called a non-existing command "jupyter baseopts" and did not complete correctly on filenames with spaces. The two patches fix these issues.


takluyver:
Thanks!

PR #131, opened 2018-05-20
 properly handle unicode characters in home dirs

vit-tucek:
Sometimes users have home directories that contain unicode characters outside of ASCII (such as č). This leads to exception thrown in `migrate.py:migrate` namely in the line:
`        dst = dst_t.format(**env)`
because `jupyter_config_dir` and `jupyter_data_dir` do not return unicode.

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -22,7 +22,8 @@ def get_home_dir():
     homedir = os.path.expanduser('~')
     # Next line will make things work even when /home/ is a symlink to
     # /usr/home as it is on FreeBSD, for example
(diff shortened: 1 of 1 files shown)


vit-tucek:
Failing because of missing `future` module. I've created another [patch](https://github.com/jupyter/jupyter_core/pull/132).

embray:
You don't need to open new PRs.  You can update the existing ones by pushing new versions of a change to the existing branch using `git push -f` at the command line.

Nevertheless, I don't think this is the correct fix.  Can you post the full traceback and exception you get when trying to run the migrate script?

vit-tucek:
Done. Now if only the devs would notice.

minrk:
Sorry for the delay on this one. Merged with a small fix.

PR #139, opened 2018-09-03
Enable system-wide installations of Jupyter possible at any path

jhgoebbert:
Extend the config paths with additional system paths from environment variable JUPYTER_CONFIG_PATH.

Motivation:
Currently Jupyter configurations are only searched at fixed system paths (sys.prefix/etc/jupyter, ...)
and jupyter_config_dir(), which is in the first place a config dir writable to the user.
For a system-wide installation of Jupyter, which is not installed at sys.prefix one needs to have the ability to add additional system config paths.

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -183,6 +183,14 @@ def jupyter_config_path():
     if os.environ.get('JUPYTER_NO_CONFIG'):
         return paths
 
+    # highest priority is env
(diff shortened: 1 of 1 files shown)


jhgoebbert:
You can find the same implementation already used for JUPYTER_PATH just a few lines above in the same file:
https://github.com/jupyter/jupyter_core/blob/master/jupyter_core/paths.py#L144

minrk:
Makes sense, thanks!

Issue #120, opened 2017-11-09
__pycache__ in the wheel


Siecje:
In the wheel on PyPI there are the following directories:
```
jupyter_core/tests/__pycache__/
jupyter_core/tests/dotipython/profile_default/__pycache__/
jupyter_core/tests/dotipython_empty/profile_default/__pycache__/
```

Should they be there?

I noticed this because we only deploy `.pyc` files so after the pip install I pull the `.pyc` files out of `__pycache__` and into the location of the `.py` files using this script.

https://bitbucket.org/Siecje/restore_pyc/src/9834a5c2c16acf29b5b330ebd7590585bc20126b/restore_pyc.py?at=default&fileviewer=file-view-default

I noticed that the resulting `.pyc` had a different magic code. There is likely a bug in the `restore_pyc.py` script linked above as it should be finding the `.pyc` file that was created during the pip install.

But I figured I would report that there are `__pycache__` directories in the wheel.

takluyver:
Not really; it's probably because of [this function](https://github.com/jupyter/jupyter_core/blob/7f1430fecbc13c2685501cff928547a92f70e2bd/setup.py#L34) which includes all directories under `tests` as package data.

Siecje:
It seems that function is not responsible, it doesn't seem to change the wheel.

If you have a `__pycache__/` directory in your package and the package is listed in `packages` then it will be included in the wheel. I guess every project has this problem.

coldfix:
This function is responsible. Via `print` debugging I found that it generates the following entries to `package_data['jupyter_core.tests']`:

```
*.*
dotipython_empty/profile_default/*.*
dotipython_empty/profile_default/static/custom/*.*
dotipython_empty/profile_default/__pycache__/*.*
__pycache__/*.*
dotipython/profile_default/*.*
dotipython/profile_default/static/custom/*.*
dotipython/profile_default/__pycache__/*.*
dotipython/nbextensions/*.*
```

The easiest workaround is to replace:

```diff
     for parent, dirs, files in os.walk(test_dir):
-        if files:
+        if files and os.path.basename(parent) != '__pycache__':
             test_files.append(pjoin(parent[prefix_len:], '*.*'))
```

as a first step.

Alternatively, make sure to build the wheel on a clean repository (e.g. `git clean -xdf`).

[...]

PR #142, opened 2018-10-17
Remove useless code

Siecje:
While investigating https://github.com/jupyter/jupyter_core/issues/120 I learned that this code is not needed. Without this code the same wheel is created.

If any of the packages contain a `__pycache__` then it will be included in the wheel.

I'm not sure how the `__pycache__` was created though. I tried installing the package and running tox and there was no `__pycache__` created.


minrk:
This isn't correct. This code is required to avoid excluding these files in source distributions (`python setup.py sdist`).

PR #147, opened 2019-03-09
Always use setuptools in setup.py

### LINKED ISSUE #120: __pycache__ in the wheel

coldfix:
Now that there are universal wheels, the setup.py is usually only ever
executed on a developer machine, where one can assume setuptools to be
available.

This slightly simplifies the setup.py code and results less potential
surprises between different setup.py runs (i.e. what happens if you run
`setup.py install` versus `setup.py bdist_wheel` and then install).

This also contains an alternate resolution for #120.

My recommendation would be to favour this PR over #146, but I can understand if you have concerns.

Best, Thomas

Resolves #120.
Closes #146.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,37 @@
+[metadata]
+name             = jupyter_core
+version          = attr: jupyter_core.version.__version__
+description      = Jupyter core package. A base package on which Jupyter projects rely.
+long_description = There is no reason to install this package on its own.
+author           = Jupyter Development Team
+author_email     = jupyter@googlegroups.org
+url              = https://jupyter.org
+license          = BSD
+classifiers      =
+    Intended Audience :: Developers
+    Intended Audience :: System Administrators
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: BSD License
+    Programming Language :: Python
+    Programming Language :: Python :: 2
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.3
+    Programming Language :: Python :: 3.4
+
+[options]
+py_modules = jupyter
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, I'm fine with this. But could we copy some bits of code that we use in other Jupyter projects to prevent setuptools from installing things as an egg:

https://github.com/jupyter/jupyter_client/blob/043c0e6a8b678efbc32d565f41d078fadddd15a5/setup.py#L47-L56

https://github.com/jupyter/jupyter_client/blob/043c0e6a8b678efbc32d565f41d078fadddd15a5/setup.py#L102-L104

Also, if we're going to move all the metadata to `setup.cfg`, let's leave a comment in `setup.py` explaining that, because people will go and look there for it, and get confused when there's just an empty `setup()` call.

coldfix:
Hi, I don't particularly like the cmdclass just to prevent users from doing `python setup.py install` or similar, but I now added it anyway as by your suggestion.

takluyver:
Thanks

PR #148, opened 2019-04-04
Removed mechanism that change $PATH

### LINKED ISSUE #145: Mutating environ PATH's side effects
Hi,

To ensure the right subcommand is launched, Jupyter is mutating the path. [https://github.com/jupyter/jupyter_core/blob/master/jupyter_core/command.py#L116](https://github.com/jupyter/jupyter_core/blob/master/jupyter_core/command.py#L116)

Can we undo this before any Kernel is launched? Otherwise the juptyer subcommands still take precedence when you do something in a subprocess from inside any kernel. This effect can be seen when you check `os.environ['PATH']` in any kernel. Aside from naming conflicts for jupyter commands, this overwrites the version of python in the path to the one you launched the jupyter server with. This is a problem when you want your path to default to python 2, but run the newest jupyter server, which requires python 3.

Please let me know if you need more info.

Thanks!

xmnlab:
resolves #145

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -225,7 +225,7 @@ def main():
         parser.print_usage(file=sys.stderr)
         sys.exit("subcommand is required")
 
-    command = _jupyter_abspath(sys.argv[0], subcommand)
+    command = _jupyter_abspath('jupyter', subcommand)
 
     try:
         _execvp(command, sys.argv[1:])
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -119,12 +119,12 @@ def _execvp(cmd, argv):
         os.execvp(cmd, argv)
 
 
-def _jupyter_abspath(path, subcommand):
+def _jupyter_abspath(subcommand):
     """This method get the abspath of jupyter with no changes on ENV."""
     # get env PATH with self
     search_path = os.pathsep.join(_path_with_self())
     # get the abs path for the jupyter-<subcommand>
-    jupyter_subcommand = '{}-{}'.format(path, subcommand)
+    jupyter_subcommand = 'jupyter-{}'.format(subcommand)
     abs_path = which(jupyter_subcommand, path=search_path)
     if abs_path is None:
         raise Exception(
@@ -225,7 +225,7 @@ def main():
         parser.print_usage(file=sys.stderr)
         sys.exit("subcommand is required")
 
-    command = _jupyter_abspath('jupyter', subcommand)
+    command = _jupyter_abspath(subcommand)
 
     try:
         _execvp(command, sys.argv[1:])
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -15,7 +15,11 @@ import os
 import sys
 from subprocess import Popen
 
-from .utils.shutil_which import which
+try:
+     # py3
+     from shutil import which
+except ImportError:
+     from .utils.shutil_which import which
 
 from . import paths
 from .version import __version__
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -120,7 +120,9 @@ def _execvp(cmd, argv):
 
 
 def _jupyter_abspath(subcommand):
-    """This method get the abspath of jupyter with no changes on ENV."""
+    """This method get the abspath of a specified jupyter-subcommand with no 
+    changes on ENV.
+    """
     # get env PATH with self
     search_path = os.pathsep.join(_path_with_self())
     # get the abs path for the jupyter-<subcommand>
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -15,19 +15,21 @@ import os
 import sys
 from subprocess import Popen
 
+from .utils.shutil_which import which
+
 from . import paths
 from .version import __version__
 
 class JupyterParser(argparse.ArgumentParser):
-    
+
     @property
     def epilog(self):
         """Add subcommands to epilog on request
-        
+
         Avoids searching PATH for subcommands unless help output is requested.
         """
         return 'Available subcommands: %s' % ' '.join(list_subcommands())
-    
+
     @epilog.setter
     def epilog(self, x):
         """Ignore epilog set in Parser.__init__"""
@@ -42,7 +44,7 @@ def jupyter_parser():
     group.add_argument('--version', action='store_true',
         help="show the jupyter command's version and exit")
     group.add_argument('subcommand', type=str, nargs='?', help='the subcommand to launch')
-    
+
     group.add_argument('--config-dir', action='store_true',
         help="show Jupyter config dir")
     group.add_argument('--data-dir', action='store_true',
@@ -53,15 +55,15 @@ def jupyter_parser():
         help="show all Jupyter paths. Add --json for machine-readable format.")
     parser.add_argument('--json', action='store_true',
         help="output paths as machine-readable json")
-    
+
     return parser
 
 
 def list_subcommands():
     """List all jupyter subcommands
-    
+
     searches PATH for `jupyter-name`
-    
+
     Returns a list of jupyter's subcommand names, without the `jupyter-` prefix.
     Nested children (e.g. jupyter-sub-subsub) are not included.
     """
@@ -89,7 +91,7 @@ def list_subcommands():
 
 def _execvp(cmd, argv):
     """execvp, except on Windows where it uses Popen
-    
+
     Python provides execvp on Windows, but its behavior is problematic (Python bug#9148).
     """
     if sys.platform.startswith('win'):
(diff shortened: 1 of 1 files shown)


xmnlab:
the error on CI for python 3.3 says: `RuntimeError: Python 3.4 or later is required`

any thoughts?

It seems it is done for review.

gnestor:
@xmnlab Your PR looks good.

It looks like master is failing CI: https://travis-ci.org/jupyter/jupyter_core/builds/443142041

@minrk Any ideas?

minrk:
We need to make sure that this doesn't change the resolution of executables, which it looks like it does, currently. I think this change currently requires *all* jupyter-foo to be in the same directory as `jupyter`, which would not be correct.

The behavior should be the same as if `dirname(jupyter)` is on PATH. So what we need to do is find `jupyter-foo` on `$PATH+dirname(jupyter)` and then resolve the absolute path, but without mutating `$PATH`. This needs to be careful to properly resolve with $PATHEXT on Windows.

One option, I believe, should be to use `shutil.which` to resolve the path:

```python
with mock.patch($PATH with self):
    abs_exe = shutil.which(jupyter-foo)
if abs_exe:
    exec(abs_exe)

**Comment by xmnlab** [comment]
thanks @minrk for the feedback. I hope to have captured your thoughts. let me know if I understood that correctly.

**Comment by xmnlab** [comment]
regarding to `https://pypi.org/project/setuptools/` it seems py3.3 is not supported any more.

**Comment by xmnlab** [comment]
it is done for a new review.

File: jupyter_core/command.py [hunk 1]
```diff
@@ -113,9 +115,24 @@ def _execvp(cmd, argv):
         os.execvp(cmd, argv)
 
 
+def _jupyter_realpath(path, subcommand):
+    # get currrent PATH
+    _path_env = os.environ['PATH']
+    # set new PATH with self
+    os.environ['PATH'] = os.pathsep.join(_path_with_self())
+    # get the real path for the jupyter-<subcommand>
+    real_path = find_executable('{}-{}'.format(path, subcommand))
```
xmnlab: shutil.which is not available for py2

xmnlab:
any feedback?

xmnlab:
hey everyone, any feedback about this PR?

xmnlab:
hey @carreau ! @gnestor told me maybe you could help with some review/feedback

xmnlab:
hey everyone! any thoughts here?

gnestor:
@takluyver Would you care to review?

takluyver:
Yup, I'll review. While we're looking at this repository, can I encourage people to consider PR #143?

jupyter_core/command.py
takluyver: But good news, this repo already has a backport - `from .utils.shutil_which import which`. This is a bit more complex than `find_executable`, e.g. it handles PATHEXT on Windows, so it would be good to use that (and the real `shutil.which` on Python 3, of course).

jupyter_core/command.py
```diff
@@ -113,9 +115,24 @@ def _execvp(cmd, argv):
         os.execvp(cmd, argv)
 
 
+def _jupyter_realpath(path, subcommand):
+    # get currrent PATH
+    _path_env = os.environ['PATH']
+    # set new PATH with self
+    os.environ['PATH'] = os.pathsep.join(_path_with_self())
```
takluyver: `shutil.which` takes a `path=` parameter to override the environment variable, so we don't even need to modify the environment variable.

jupyter_core/command.py
```diff
@@ -113,9 +115,24 @@ def _execvp(cmd, argv):
...
+    # get the real path for the jupyter-<subcommand>
+    real_path = find_executable('{}-{}'.format(path, subcommand))
+    # RESTORE previous PATH
+    os.environ['PATH'] = _path_env
+    return (
+        real_path if real_path is not None and os.access(real_path, os.X_OK)
+        else None
```
takluyver: We'll probably get more meaningful error messages if we throw an error when we can't find the command rather than returning None.

jupyter_core/command.py
```diff
@@ -113,9 +115,24 @@ def _execvp(cmd, argv):
         os.execvp(cmd, argv)
 
 
+def _jupyter_realpath(path, subcommand):
```
takluyver: This should have a brief docstring to explain what it is.

xmnlab:
thanks for the feedback @takluyver i will work on that!

xmnlab:
@takluyver  @minrk @gnestor it is done for a new review :)

jupyter_core/command.py
```diff
@@ -15,19 +15,21 @@
 import sys
 from subprocess import Popen
 
+from .utils.shutil_which import which
```
takluyver: We should use the stdlib implementation where it's available, like this:

https://github.com/jupyter/jupyter_core/blob/0fe5994c2ec528de4cb01178a296d95db945fa06/jupyter_core/application.py#L18-L22

jupyter_core/command.py
```diff
@@ -197,12 +216,13 @@ def main():
...
             return
-    
+
     if not subcommand:
         parser.print_usage(file=sys.stderr)
         sys.exit("subcommand is required")
-    
-    command = 'jupyter-' + subcommand
+
+    command = _jupyter_abspath(sys.argv[0], subcommand)
```
takluyver: I think we always want to be looking up `jupyter-xyz`, regardless of what `sys.argv[0]` is.

xmnlab: @takluyver I changed that considering the follow case reported by @mlucool 

```
Setup:
1. Set your PATH to have python 2 as your default version of python (prepend the path)
2. Run jupyterlab from a python 3 location (e.g. `path/to/python-3/bin/jupyterlab')
3. In a python 2 kernel run the following:
```
import sys
print(sys.version)
import os
print(os.environ['PATH'])
```
As expected, sys.version will be python 2. But now you can see that the PATH is modified with python3 in preprended. So if I were to do:
```
import subprocess
print(subprocess.Popen("python -c 'import sys; print(sys.version)'", shell=True, stdout=subprocess.PIPE).stdout.read())
```
This would show python 3.

This is really just a problem as we are migrating from python 2 to 3 and we want to run the server with 3 (as we want to use latest) but anything that is called via subprocess should use whatever you had in your PATH. We are planning to locally path `ipykernel_launcher.py` in a hacky way. This is suboptiomal so we'd like to revert this as soon as we can.
```

maybe it will not be a problem considering that there is no jupyterlab for py2 .. but if `path/to/python-3/bin/jupyterlab` is not in `$PATH` even if I run `jupyerlab` using the full path it will not be found.

do you have any thoughts here, considering this case?

takluyver: I'm fine with not modifying `$PATH`.

The `sys.argv[0]` that you pass in here is only used, as far as I can see, in this line:

```python
jupyter_subcommand = '{}-{}'.format(path, subcommand)
```

If `sys.argv[0]` is, e.g. `/home/takluyver/.local/bin/jupyter`, then you construct a subcommand like `/home/takluyver/.local/bin/jupyter-notebook`. That short-circuits `shutil.which()`,  so it will only work if that's the location of `jupyter-notebook`.

If the subcommand is `notebook`, then we should always be looking up `jupyter-notebook`, not `{sys.argv[0]}-notebook`.

xmnlab: I see .. OK .. so maybe related to the scenario presented by @mlucool the workaround should be call `/path/jupyterlab-subcommand` directly

xmnlab: @takluyver it is done for a new review! thanks!

jupyter_core/command.py
```diff
@@ -113,9 +119,29 @@ def _execvp(cmd, argv):
         os.execvp(cmd, argv)
 
 
+def _jupyter_abspath(path, subcommand):
+    """This method get the abspath of jupyter with no changes on ENV."""
+    # get env PATH with self
+    search_path = os.pathsep.join(_path_with_self())
+    # get the abs path for the jupyter-<subcommand>
+    jupyter_subcommand = '{}-{}'.format(path, subcommand)
```
takluyver: Thanks. Now you don't need to pass an argument, you can just hardcode it in this line. That's one less thing to follow around when reading the code. :-)

xmnlab: Ok :) done!

jupyter_core/command.py
```diff
@@ -113,9 +119,29 @@ def _execvp(cmd, argv):
         os.execvp(cmd, argv)
 
 
+def _jupyter_abspath(subcommand):
+    """This method get the abspath of jupyter with no changes on ENV."""
```
takluyver: Nitpick: "...the abspath of a specified jupyter-subcommand..."

takluyver:
Thanks! I've just spotted that the docstring should be a bit clearer. Other than that, I think this looks good.

@gnestor @minrk - once this is landed, I'll probably try to make a 4.5.0 release soon.

jupyter_core/command.py
xmnlab: ok! done! thanks for the feedback!

PR #149, opened 2019-04-12
Respect the raise_config_file_errors trait and re-raise if applicable.

ian-r-rose:
I was taking a look at jupyterlab/jupyterlab#5716, and found that setting the base `Application` trait `raise_config_file_errors` wasn't behaving as I expected. This is because `JupyterApp` overloads the `load_config_file` function and doesn't check this option.

A quick audit of the Jupyter ecosystem shows that there are a couple of different behaviors around this.
* [`jupyterhub`](https://github.com/jupyterhub/jupyterhub/commit/7968912a7c6dda77472af15230bf6ae40e3324b2) inherits from `Application` rather than `JupyterApp`, so the trait works as expected.
* [`IPython`](https://github.com/ipython/traitlets/pull/228) overrides `load_config_file`, and specifically checks for this option.
* Notebook and JupyterLab inherit from `JupyterApp`, and they do not respect this option.

Another option would be to override the `load_config_file` option in lab/notebook, but it seemed cleaner to me to add this check here.

### DIFF
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -203,8 +203,9 @@ class JupyterApp(Application):
         except ConfigFileNotFound:
             self.log.debug("Config file not found, skipping: %s", config_file_name)
         except Exception:
-            # For testing purposes.
-            if not suppress_errors:
+            # Reraise errors for testing purposes, or if set in
+            # self.raise_config_file_errors
+            if not suppress_errors or self.raise_config_file_errors:
                 raise
             self.log.warning("Error loading config file: %s" %
                             config_file_name, exc_info=True)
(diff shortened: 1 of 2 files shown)


ian-r-rose:
Failure is due to setuptools no longer supporting python 3.3 (as of June 2018).

ian-r-rose:
Ping @takluyver, any thoughts on this?

jupyter_core/application.py
```diff
@@ -203,8 +203,9 @@ def load_config_file(self, suppress_errors=True):
         except ConfigFileNotFound:
             self.log.debug("Config file not found, skipping: %s", config_file_name)
         except Exception:
-            # For testing purposes.
-            if not suppress_errors:
+            # Reraise errors for testing purposes, or if set in
+            # self.raise_config_file_errors
+            if not suppress_errors or self.raise_config_file_errors:
```
takluyver: Can we put `not suppress_errors` in parentheses? I never remember operator precedence orders.

takluyver:
Thanks

ian-r-rose:
Thanks for the review + merge!

Issue #160, opened 2019-08-22
Feature request: jupyter command searches in Python's "scripts" directory for subcommands


pfmoore:
My use case here is fairly unusual, so it may be that this is something you don't want to support. But what I am trying to do is as follows. I want to keep my Jupyter environment isolated from my system Python, so I install it into a virtual environment. If I do that, and run `jupyter notebook`, either with the venv activated, or using the full path to the `jupyter.exe` command, everything works as expected.

However, I would rather not need to activate the environment, or use the full path (I don't want to accidentally run the venv python, and typing the full path on the command line is a pain). I could use an alias for that, but what I usually do is copy the script wrapper (`jupyter.exe`) to my `~/.local/bin` directory. This works for other applications, but jupyter fails to find its subcommands if I do this.

The problem seems to be in `jupyter_core.command` in function `_path_with_self`, which adds the directory containing the executable to `PATH`, for looking up subcommands. That fails if the `jupyter.exe` program gets moved.

A fix which seems to work for me is to add `sysconfig.get_path('scripts')` to the list of paths added to `PATH`. A proof of concept fix to `_path_with_self` is as follows:

```python
def _path_with_self():
    """Put `jupyter`'s dir at the front of PATH

    Ensures that /path/to/jupyter subcommand
    will do /path/to/jupyter-subcommand
    even if /other/jupyter-subcommand is ahead of it on PATH
    """
    path_list = (os.environ.get('PATH') or os.defpath).split(os.pathsep)

    # Insert the "scripts" directory for this Python installation
    import sysconfig
    path_list.insert(0, sysconfig.get_path('scripts'))

    scripts = [sys.argv[0]]
    # Rest is unchanged
```

Is this something that could be considered for inclusion? I don't think it will cause problems anywhere else, and even though my use case is unusual, looking in the `scripts` location is a fairly natural and reasonable thing to do anyway.

I can create a PR for this if there is interest.

EFanZh:
I am using Python 3.8 installed from [Microsoft Store](https://www.microsoft.com/en-us/p/python-38/9mssztt1n39l). I noticed that the directory returned by `sysconfig.get_path('scripts')` does not exist, but the directory returned by `sysconfig.get_path('scripts', 'nt_user')` exists, and with proper executables in it. Since `sysconfig.get_path('scripts')` doesn’t exist, I couldn’t launch Jupyter Notebook.

pfmoore:
[...]

PR #162, opened 2019-09-09
Add the environment scripts directory to the list of places to search for subcommands

pfmoore:
Implementation of #160


minrk:
Makes sense, thanks!

PR #163, opened 2019-09-13
Moved path util code from jupyter_client and jupyter_server to core

MSeal:
Moved shared code from https://github.com/jupyter/jupyter_server/pull/77 and https://github.com/jupyter/jupyter_client/pull/469 as well as some other path related utils in jupyrer_server that seemed useful in a more generic scope. No code changes made on the functions  other than converting the tests from nose.assert pattern to pytest assert pattern.

Tested on windows & ubuntu independently

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -11,9 +11,18 @@
 
 import os
 import sys
+import stat
+import errno
 import tempfile
 
+from contextlib import contextmanager
+from ipython_genutils import py3compat
+
 pjoin = os.path.join
+
+# UF_HIDDEN is a stat flag not defined in the stat module.
+# It is used by BSD to indicate hidden files.
+UF_HIDDEN = getattr(stat, 'UF_HIDDEN', 32768)
 
 
 def get_home_dir():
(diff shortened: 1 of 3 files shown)


jupyter_core/paths.py
```diff
@@ -187,3 +196,213 @@ def jupyter_config_path():
...
+        # Touch file pre-emptively to avoid editing permissions in open files in Windows
+        fd = os.open(fname, os.O_CREAT | os.O_WRONLY | os.O_TRUNC, 0o600)
+        os.close(fd)
+        open_flag = os.O_WRONLY | os.O_TRUNC
+        win32_restrict_file_to_user(fname)
+
+    with os.fdopen(os.open(fname, open_flag, 0o600), mode) as f:
+        if os.name != 'nt':
+            # Enforce that the file got the requested permissions before writing
+            assert '0600' == oct(stat.S_IMODE(os.stat(fname).st_mode)).replace('0o', '0')
```
vidartf: Do we really not care for this check if running in optimized mode (since it uses `assert` instead of `if`/`raise`)?

MSeal: It should probably be a raise and not an assert, though it's not critical to do this check. I'll make the change tomorrow.

vidartf: Sure, I was just linting a bit while waiting for someone else to have a look at it. The comment did say "enforce", which made it sound a little more critical 🤷‍♂

MSeal: Np, this is what code reviews are for. I've updated the assertion check.

minrk:
Makes sense, thanks

Issue #168, opened 2019-09-30
jupyter bash completion OS X compatibility


alexisdimi:
The current version of jupyter bash completion (`examples/jupyter-completion.bash`, see [Issue #121](https://github.com/jupyter/jupyter_core/issues/121)), uses a bash 4.x feature and will not work on OS X, which comes with bash version 3.2. 

**Resolution:** It is very easy to fix this and achieve bash 3.2 compatibility: Just rewrite these two lines near the end of the file, from

    <        compopt -o filenames
    <        COMPREPLY=( $(compgen -f -- ${cur}) )
to simply

    >        COMPREPLY=( $(compgen -o filenames -f -- ${cur}) )

This achieves precisely the same behavior without involving the troublesome command `compopt`.

willingc:
Closed by #169

PR #169, opened 2019-10-03
Update example to achieve bash 3.2 compatibility

themarkrogers:
as per alexisdimi's suggestion in Issue #168


rgbkrk:
If only we could update MacOS to Bash 4.x 😄

PR #176, opened 2019-10-27
Travis CI: Add Python 3.7 and 3.8 and remove the EOL 3.4

cclauss:
Travis runtimes are interesting.  Why does `pip install -r requirements.txt` take several times longer on Py38 and nightly than all other Python versions? https://travis-ci.org/jupyter/jupyter_core/builds/603413767?utm_source=github_status&utm_medium=notification

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 # http://travis-ci.org/#!/jupyter/jupyter_core
 language: python
-sudo: false
 python:
   - nightly
+  - "3.8"
+  - "3.7"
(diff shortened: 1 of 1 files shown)


takluyver:
It probably has to compile something from source which is available as a pre-built wheel for the other versions of Python. Looking at the log, pyzmq and tornado both get built as wheels with a `linux_x86_64` tag, which means they have compiled parts. This should resolve itself as those projects make new releases with Python 3.8 wheels.

PR #177, opened 2019-11-01
Ignore sticky bit when validating permissions.

WibbletheDuck:
On [Nextjournal](https://nextjournal.com/) we use a standard connection file which is mounted to docker containers in order to support Jupyter runtimes.  Since the addition of `secure_write` we can get an initial startup just fine by setting permissions on the file.  However, with some Jupyter kernels it seems the Sticky Bit gets set, and this persists on the mounted file through restarts and resets, causing `secure_write` to fail:
```
RuntimeError: Permissions assignment failed for secure file: '/runtimes/jupyter/kernel_connection.json'.Got '0o1600' instead of '0o0600'
```
I don't think the Sticky Bit is a security risk(?), so I see no reason it shouldn't be ignored.  [Tested the fork](https://nextjournal.com/a/LnBMKtooyHaE7id2qrbrG?token=RjwFVRya7ycP1uQ3E9xCyw), and this change does clear up our problem.

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -388,7 +388,7 @@ def get_file_mode(fname):
     # Some filesystems (e.g., CIFS) auto-enable the execute bit on files.  As a result, we
     # should tolerate the execute bit on the file's owner when validating permissions - thus
     # the missing one's bit on the third octet.
-    return stat.S_IMODE(os.stat(fname).st_mode) & 0o7677  # Use 4 octets since S_IMODE does the same
+    return stat.S_IMODE(os.stat(fname).st_mode) & 0o6677  # Use 4 octets since S_IMODE does the same
 
 
 @contextmanager
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -386,8 +386,9 @@ def get_file_mode(fname):
 
     """
     # Some filesystems (e.g., CIFS) auto-enable the execute bit on files.  As a result, we
-    # should tolerate the execute bit on the file's owner when validating permissions - thus
-    # the missing one's bit on the third octet.
+    # should tolerate the execute bit on the file's owner when validating permissions - thus 
+    # the missing least significant bit on the third octet. In addition, we also tolerate 
+    # the sticky bit being set, so the lsb from the fourth octet is also removed.
     return stat.S_IMODE(os.stat(fname).st_mode) & 0o6677  # Use 4 octets since S_IMODE does the same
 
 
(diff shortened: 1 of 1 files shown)


jupyter_core/paths.py
```diff
@@ -388,7 +388,7 @@ def get_file_mode(fname):
     # Some filesystems (e.g., CIFS) auto-enable the execute bit on files.  As a result, we
     # should tolerate the execute bit on the file's owner when validating permissions - thus
     # the missing one's bit on the third octet.
-    return stat.S_IMODE(os.stat(fname).st_mode) & 0o7677  # Use 4 octets since S_IMODE does the same
+    return stat.S_IMODE(os.stat(fname).st_mode) & 0o6677  # Use 4 octets since S_IMODE does the same
```
kevin-bates: I agree we should ignore the sticky bit.  Wondering if we shouldn't just ignore them all in that octet,  but that's a larger discussion.

If we're only removing the sticky bit from the mask, then the comment should be updated to state why.  I also had a brain cramp when I added the previous comment and we should change the wording of "_one's bit_" to "_least significant bit_" while we're here.  So perhaps something like...

> ...thus the missing least significant bit on the third octet.  In addition, we also tolerate the sticky bit being set, so the lsb from the fourth octet is also removed.

The other "repair" ask I have (sorry), is to fix the error message at the [end of file](https://github.com/jupyter/jupyter_core/pull/177/files#diff-6cd671a3971189df64cfe37964c2047cR432).  Could you please add some spacing before "Got" and put a period at the end of its sentence as well?  That's been bugging me in the recent logs I've seen.  :smile:

WibbletheDuck: 👍 ?

takluyver: I'm going to go super nitpicky about the comments: I think we're using 'octet' inaccurately. An [octet](https://en.wikipedia.org/wiki/Octet_(computing)) is 8 bits, aka a byte. Digits in octal represent 3 bits each. So I think we should be using the term 'octal digit' instead.

(I know this was already there, but I think it's worth fixing while we're looking at this bit of code anyway)

kevin-bates: Yeah, you're right - my bad.  @WibbletheDuck - would you mind fixing up the occurrences of `octet` as well?

WibbletheDuck: @kevin-bates Done!

kevin-bates: Thank you!

takluyver:
Thanks

PR #178, opened 2019-11-08
Add readthedocs YAML config

takluyver:
Doing `setup.py install` is blocked to work around setuptools problems, and doing `pip install .` instead can seemingly only be configured through the YAML file.


takluyver:
Merging this because I want to try to fix the docs build while I still remember it.

takluyver:
Docs are building again.

Issue #185, opened 2020-04-03
Add support for XDG_CONFIG_HOME


0az:
Right now, I have a patch (wip) to implement `XDG_CONFIG_HOME` on an opt-in basis. That is, `XDG_CONFIG_HOME` is used *if and only if* `.jupyter` does not exist. This minimizes breakage by ensuring that the change only affects:

1. New users with `XDG_CONFIG_HOME` set,
2. Existing users with `XDG_CONFIG_HOME` set who explicitly migrate,
3. Existing users with `XDG_CONFIG_HOME` set who delete their `.jupyter` directory for any other reason.

The solution to cases 1/3 is documentation, and I'm willing to update the docs in jupyter/jupyter, and any other locations.

The question is: should the Linux default get set to `.config`, when it isn't set? This is a more significant change in behavior, but it parallels the existing default of `.local/share`.

jasongrout:
I think this is now implemented with #292, where we use xdg if it is defined.

PR #186, opened 2020-04-04
Add opt-in support for XDG_CONFIG_HOME, XDG_DATA_HOME

0az:
- Allow opt-in use of `XDG_CONFIG_HOME` on all platforms
- Allow opt-in use of `XDG_DATA_HOME` on Darwin

Related: #185


nfultz:
This would be nice to have.

PR #190, opened 2020-05-22
Avoid using ctypes on recent Python versions

zooba:
This check can be handled with core functionality, so the ctypes dependency is redundant.

I haven't kept up on the actual supported versions right now, but if everything before 3.5 is not covered anymore, the fallback code (`if not hasattr`) can be removed completely.


MSeal:
Sorry I didn't noticed this PR before -- but we're not supporting python 2 anymore so I think we can further simplify this PR and remove the ctypes check entirely.

jasongrout:
Python 3.5 also has reached EOL: https://www.python.org/downloads/release/python-350/

jasongrout:
> I haven't kept up on the actual supported versions right now, but if everything before 3.5 is not covered anymore, the fallback code (`if not hasattr`) can be removed completely

I added a commit here that did this.

I also slightly restructured the logic to mirror the logic in the corresponding posix function for consistency.

jasongrout:
Rebased on #201 - please review that one first, as it converts our testing over to github actions.

PR #195, opened 2020-08-17
Remove python 2 compatibility from package

rahulporuri:
Assuming that going forward, the plan is to drop python 2 compatibility, this PR updates all of the python 2/3 compatibility code in this package to remove support for python 2.

- remove `coding:` cookies from files if the encoding is utf-8 as utf-8 is the default file encoding on python 3
- remove `jupyter_core.utils.shutil_which` and replace it with `shutil.which`
- remove `__future__` imports
- remove `try/except` blocks for unittest.mock imports
- remove uses of `ipython_genutils.py3compat`
- replace `raw_input` with `input`


rahulporuri:
@MSeal will you be able to review this PR? (I'm not sure who the maintainers for this package are)

Issue #193, opened 2020-08-10
Use of distutils triggers warning when setuptools (>=v49.2.0) is used in jupyter kernels


mstimberg:
Since version v49.2.0, `setuptools` is raising a warning if it gets imported after `distutils`. Unfortunately, `jupyter_code` uses the `strtobool` function from `distutils` in https://github.com/jupyter/jupyter_core/blob/ed50a593d03684d93c0234372cd2947e500ef7b4/jupyter_core/paths.py#L397
which means that any jupyter notebook/console that imports `setuptools` (or a package that uses `setuptools`) will show this warning. This is very confusing, especially for beginners who often mistake this warning for an error. 

I don't think that `jupyter_core` is to blame here, but maybe it would be worth getting rid of this `distutil` use? It seems that this warning might only be around until the next release (see discussion in pypa/setuptools#2259), though, so it is rather a temporary issue that might go away soon.

pganssle:
Regardless of the erroneous warning about the import order (which affects nothing), you should definitely remove all use of `distutils` that it is plausible to use. The erroneous warnings are only the first stage of a larger effort to deprecate and remove `distutils` itself. Replacements are not available for everything in `distutils`, but I doubt that we'll ever do anything except remove `strtobool`, since that's hardly in the core API of a build tool.

MSeal:
There's not many people actively maintaining some of these libraries (I'm certainly stretched too thin), but I'd happily merge a change if someone could take the time to make the appropriate deprecation removals.

PR #197, opened 2020-09-07
Replace strtobool to avoid distutils import

mstimberg:
This removes the `strtobool` import from `distutils.util`, used for evaluating the value of the `JUPYTER_ALLOW_INSECURE_WRITES` environment variable, and replaces it by a straightforward expression. See comments in #193 for the justification.

Strictly speaking, this breaks backwards compatibility because the proposed implementation:
1. only accepts "true" (ignoring case) and "1" as `True` values, whereas the previous implementation also accepted "y", "yes", "t", and "on".
2. stores a `True`/`False` boolean in the `allow_insecure_writes` variable, whereas the previous implementation stored integer numbers `1`/`0`.

To me (as an outsider of the jupyter project), this does not seem to be a problem since these implementation details were only available from the source code, and internally the variable is only used as a boolean value. As far as I can tell, the only public documentation of the feature is [here](https://discourse.jupyter.org/t/jupyter-core-4-6-2-release-with-insure-mode-option/3300) and only mentions "true" and "1". All other mentions I found (e.g. on [stackoverflow](https://stackoverflow.com/questions/59934047/jupyter-access-denied-error-when-starting-kernel-in-spyder-jupyter-notebook)) seem to use "1" as a value.

I did not add any tests, since I think you'd need to restructure the code a bit for proper testing (the `allow_insecure_writes` variable is currently set at module import time).

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -18,7 +18,6 @@ import warnings
 from ipython_genutils import py3compat
 
 from contextlib import contextmanager
-from distutils.util import strtobool
 from ipython_genutils import py3compat
 
 pjoin = os.path.join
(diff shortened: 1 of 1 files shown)


kevin-bates:
I had happened to reference this yesterday in [discourse](https://discourse.jupyter.org/t/kernel-error-in-jupyter-notebook/4728/3?u=kevin-bates) but fortunately used 'true'.  I have just added `(or 1)` as well.

I'm fine with this change and it seems like a good, proactive, approach.  Thanks for raising the issue and providing the pull request.

kevin-bates (approved):
Looks good - thanks Marcel.

MSeal (approved):

MSeal:
Thanks @mstimberg !

PR #199, opened 2020-10-29
Add an environment variable flag to switch the user and environment path order

jasongrout:
This addresses the long-standing problem we have when running Jupyter in virtual environments, where a given user will have many virtual environments. In this case, conceptually the environment-level paths are more specific than user-level paths, but Jupyter still treats the user-level paths as more specific. If a user typically uses virtual environments, they can set this flag in their environment and Jupyter will prioritize environment-level paths over user-level paths.

So now, setting `JUPYTER_ENV_PRIORITY` environment variable to anything except 'n', 'no', 'false', 'off', '0', or '0.0' makes Jupyter use environment paths first, *then* user-level paths (including JUPYTER_DATA_DIR or JUPYTER_CONFIG_DIR, if specified), then system locations.

We also fix an inconsistency between JUPYTER_PATH and JUPYTER_CONFIG_PATH, so that JUPYTER_CONFIG_PATH now takes priority over the user-level config directory, as happens for JUPYTER_PATH.

Questions about implementation:
1. The environment variable name is JUPYTER_ENV_PRIORITY. I don't particularly like this name. Any other suggestions?
2. ~~To determine if the variable is set (as a boolean flag), I use `bool(os.environ.get('JUPYTER_ENV_PRIORITY'))`. Is this a normal convention? For example, any nonempty string will act as true (even "false" or "False"), but the empty string will count as false. This seems a bit confusing. How should environment variables that are binary flags work?~~ *Edit: I fixed this in a later commit: False is case insensitive no|n|false|off|0|0.0, and true is anything else.*

- [x] Add tests

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -24,7 +24,7 @@ pjoin = os.path.join
 # It is used by BSD to indicate hidden files.
 UF_HIDDEN = getattr(stat, 'UF_HIDDEN', 32768)
 
-JUPYTER_ENV_PRIORITY = bool(os.environ.get('JUPYTER_ENV_PRIORITY'))
+JUPYTER_ENV_PRIORITY = os.environ.get('JUPYTER_ENV_PRIORITY', 'n').lower() not in ['no', 'n', 'false', 'off', '0', '0.0']
 
 
 def get_home_dir():
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -190,8 +190,8 @@ ENV_CONFIG_PATH = [os.path.join(sys.prefix, 'etc', 'jupyter')]
 def jupyter_config_path():
     """Return the search path for Jupyter config files as a list.
     
-    If JUPYTER_ENV_PRIORITY is set, the environment-level directories
-    will have priority over user-level directories.
+    If the JUPYTER_ENV_PRIORITY environment variable is set, the environment-level
+    directories will have priority over user-level directories.
     """
     if os.environ.get('JUPYTER_NO_CONFIG'):
         return [jupyter_config_dir()]
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -24,8 +24,9 @@ pjoin = os.path.join
 # It is used by BSD to indicate hidden files.
 UF_HIDDEN = getattr(stat, 'UF_HIDDEN', 32768)
 
-JUPYTER_ENV_PRIORITY = os.environ.get('JUPYTER_ENV_PRIORITY', 'n').lower() not in ['no', 'n', 'false', 'off', '0', '0.0']
 
+# True if environment variable is set to anything besides no, n, false, off, 0, or 0.0 (case insensitive)
+JUPYTER_PREFER_ENV_PATH = os.environ.get('JUPYTER_PREFER_ENV_PATH', 'n').lower() not in ['no', 'n', 'false', 'off', '0', '0.0']
 
 def get_home_dir():
     """Get the real path of the home directory"""
@@ -132,7 +133,7 @@ def jupyter_path(*subdirs):
 
     JUPYTER_PATH environment variable has highest priority.
 
-    If the JUPYTER_ENV_PRIORITY environment variable is set, the environment-level
+    If the JUPYTER_PREFER_ENV_PATH environment variable is set, the environment-level
     directories will have priority over user-level directories.
 
     If ``*subdirs`` are given, that subdirectory will be added to each element.
@@ -156,7 +157,7 @@ def jupyter_path(*subdirs):
     user = jupyter_data_dir()
     env = [p for p in ENV_JUPYTER_PATH if p not in SYSTEM_JUPYTER_PATH]
 
-    if JUPYTER_ENV_PRIORITY:
+    if JUPYTER_PREFER_ENV_PATH:
         paths.extend(env)
         paths.append(user)
     else:
@@ -190,7 +191,7 @@ ENV_CONFIG_PATH = [os.path.join(sys.prefix, 'etc', 'jupyter')]
 def jupyter_config_path():
     """Return the search path for Jupyter config files as a list.
     
-    If the JUPYTER_ENV_PRIORITY environment variable is set, the environment-level
+    If the JUPYTER_PREFER_ENV_PATH environment variable is set, the environment-level
     directories will have priority over user-level directories.
     """
     if os.environ.get('JUPYTER_NO_CONFIG'):
@@ -207,7 +208,7 @@ def jupyter_config_path():
     user = jupyter_config_dir()
     env = [p for p in ENV_CONFIG_PATH if p not in SYSTEM_CONFIG_PATH]
 
-    if JUPYTER_ENV_PRIORITY:
+    if JUPYTER_PREFER_ENV_PATH:
         paths.extend(env)
         paths.append(user)
     else:
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -25,8 +25,13 @@ pjoin = os.path.join
 UF_HIDDEN = getattr(stat, 'UF_HIDDEN', 32768)
 
 
-# True if environment variable is set to anything besides no, n, false, off, 0, or 0.0 (case insensitive)
-JUPYTER_PREFER_ENV_PATH = os.environ.get('JUPYTER_PREFER_ENV_PATH', 'n').lower() not in ['no', 'n', 'false', 'off', '0', '0.0']
+def envset(name):
+    """Return True if the given environment variable is set
+
+    An environment variable is considered set if it is assigned to a value
+    other than 'no', 'n', 'false', 'off', '0', or '0.0' (case insensitive)
+    """
+    return os.environ.get(name, 'no').lower() not in ['no', 'n', 'false', 'off', '0', '0.0']
 
 def get_home_dir():
     """Get the real path of the home directory"""
@@ -58,7 +63,7 @@ def jupyter_config_dir():
     env = os.environ
     home_dir = get_home_dir()
 
-    if env.get('JUPYTER_NO_CONFIG'):
+    if envset('JUPYTER_NO_CONFIG'):
         return _mkdtemp_once('jupyter-clean-cfg')
 
     if env.get('JUPYTER_CONFIG_DIR'):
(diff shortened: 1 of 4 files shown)


jasongrout:
As for boolean environment variables, here are some conventions:

* true values are 'true', 'on', 'yes' and all numbers except 0 and 0.0; everything else is false (https://symfony.com/doc/current/configuration/env_var_processors.html)
* False values are false, f, 0, off, everything else is true (https://api.rubyonrails.org/classes/ActiveModel/Type/Boolean.html)
* values  y|Y|yes|Yes|YES|n|N|no|No|NO|true|True|TRUE|false|False|FALSE|on|On|ON|off|Off|OFF (https://yaml.org/type/bool.html)

jasongrout:
CCing a few people for review:

@blink1073 
@SylvainCorlay - remember how we discussed doing something like this years and years ago? Well, here it is :)
@mseal

jasongrout:
Rebased on #201 - please review that one first, as it converts our testing over to github actions.

jupyter_core/paths.py
```diff
@@ -24,14 +24,15 @@
 # It is used by BSD to indicate hidden files.
 UF_HIDDEN = getattr(stat, 'UF_HIDDEN', 32768)
 
+JUPYTER_ENV_PRIORITY = os.environ.get('JUPYTER_ENV_PRIORITY', 'n').lower() not in ['no', 'n', 'false', 'off', '0', '0.0']
```
kevin-bates, line 27: Regarding the name, I think it should include `PATH` in some manner, otherwise its a bit ambiguous as readers might wonder 'what environment is being prioritized'.

Some suggestions:
`JUPYTER_PREFER_ENV_PATH`
`JUPYTER_ENV_PATH_PRIORITY`
`JUPYTER_PRIORITIZE_ENV_PATH`

or using `PYENV` over `ENV`, then `JUPYTER_PYENV_PRIORITY` isn't so ambiguous either.

A short comment might be helpful here as well.

jupyter_core/paths.py
```diff
@@ -131,6 +132,9 @@ def jupyter_path(*subdirs):
 
     JUPYTER_PATH environment variable has highest priority.
 
+    If the JUPYTER_ENV_PRIORITY environment variable is set, the environment-level
```
kevin-bates, line 135: Just wondering if there might be a case where, say, a server extension might require the current priority model, yet needs to protect itself against the new env, or vice versa (i.e., wants to force the new behavior for its functionality)?

To support that, we could add an optional kwarg (e.g., `env_priority=None`) such that, when set, it takes precedence over the ENV value.  (The argument name should be that of env name sans the `jupyter_` prefix.)

jupyter_core/paths.py
jasongrout: or `JUPYTER_PATH_PREFER_ENV` so it is sort of next to `JUPYTER_PATH`, which is related (though this also applies to `JUPYTER_CONFIG_PATH`, so perhaps `JUPYTER_PREFER_ENV`)?

I like PREFER over PRIORITY as it is a simpler word.

jasongrout: `JUPYTER_PREFER_ENV`?

jasongrout: On second thought, I like `JUPYTER_PREFER_ENV_PATH` best of the options suggested so far. `JUPYTER_PATH_*` makes it seem like it doesn't apply to jupyter config, whereas `JUPYTER_PREFER_ENV_PATH` makes it clearer that we are talking about the environment path (where it comes up) rather than Jupyter's path.

jupyter_core/paths.py
jasongrout: I'm having trouble coming up with a case where you want to override the user's explicit wishes that the environment be given priority over the user level. Can you elaborate on this case?

I see this variable as the user giving Jupyter explicit information about a common usecase that was not envisioned when originally coding this path hierarchy.

jasongrout: Also, I kind of hesitate to introduce an override of the override - that is getting pretty confusing unless we really have a concrete case for it, I think.

kevin-bates: Understood. I'm okay with that.  It's something that could be added at any time (following the existence of a concrete case).

jupyter_core/paths.py
jasongrout: Fixed in latest commit.

kevin-bates:
This looks good Jason.  One thing I'm wondering if whether we should log an informational message (once only) when the override is in play as I could see there being confusion during troubleshooting, especially since `jupyter --paths` could be misleading if/when invoked from a shell that has a different env value from the application's env.

It may not be a bad idea that that same info message get posted in `jupyter --paths` - which might provide the troubleshooters a hint one way or another (and given the application info message may be long gone by then).

jasongrout:
Looking through things one last time, I realized we have an inconsistency. Here we use an environment variable as a boolean flag, but we only check to see if it is set or not (ignoring any value it may have): https://github.com/jupyter/jupyter_core/pull/199#diff-e185ceb5aea025405e24e7c43c570a827b66ff50b33e43a99f2d7c66ea192316R199

Whereas the new code I added checks not only if it is set, but also if some common values are set like 'n', 'no', 'off', 'false', '0', '0.0': https://github.com/jupyter/jupyter_core/pull/199#diff-e185ceb5aea025405e24e7c43c570a827b66ff50b33e43a99f2d7c66ea192316R29

We should probably be consistent here with our environment variable boolean flags. Changing the behavior of `JUPYTER_NO_CONFIG` would be a breaking change. Should we change `JUPYTER_PREFER_ENV_PATH` to just check to see if the environment variable is set, but completely ignore the value, so `JUPYTER_PREFER_ENV_PATH=false` would still count as preferring the environment path?

jasongrout:
>  One thing I'm wondering if whether we should log an informational message (once only)

Not sure how we would do once only (per what?) cleanly.



> It may not be a bad idea that that same info message get posted in `jupyter --paths` - which might provide the troubleshooters a hint one way or another (and given the application info message may be long gone by then).

Sure, I can add an info log.

jasongrout:
On the other hand, we don't log anything when JUPYTER_NO_CONFIG, JUPYTER_PATH, JUPYTER_CONFIG_PATH, JUPYTER_CONFIG_DIR, or JUPYTER_DATA_DIR are set, which all influence these paths too. Seeing that, and that there are no other logging messages generated, probably it's not worth logging this?

kevin-bates:
> Not sure how we would do once only (per what?) cleanly.

I was thinking once per process since I suspect it could get noisy with long-running applications.  You'd probably need a global although I'll admit I'm not up on how python loading works completely.  Perhaps the approach taken in `jupyter --paths` will shed some light on this?

> We should probably be consistent here with our environment variable boolean flags. Changing the behavior of JUPYTER_NO_CONFIG would be a breaking change. Should we change JUPYTER_PREFER_ENV_PATH to just check to see if the environment variable is set, but completely ignore the value so it is consistent, so JUPYTER_PREFER_ENV_PATH=false would still count as preferring the environment path?

This is unfortunate, but believe we _could_ make this change (to honor appropriate values) since folks using `JUPYTER_NO_CONFIG=0|false` and expecting their configuration lookups to be bypassed are living on the edge in the first place.  Since you have coded explicit "false" values, it would only be "breaking" those users that, well, some might say, deserve to break.   As such, I'd leave things how you have them and don't really see an issue updating `JUPYTER_NO_CONFIG` now or at a later time (perhaps in a minor or major release).

jasongrout:
> As such, I'd leave things how you have them and don't really see an issue updating `JUPYTER_NO_CONFIG` now or at a later time (perhaps in a minor or major release).

Sounds good, and we can update JUPYTER_NO_CONFIG when there is a major release.

The release containing this recent spate of PRs - minor or major? I think it could be minor.

kevin-bates:
> The release containing this recent spate of PRs - minor or major? I think it could be minor.

I agree, minor seems fine.

jasongrout:
I pushed a change that adds a `--debug` switch to `jupyter --paths`

jasongrout:
I rebased on master to clean up the commit list.

kevin-bates:
Hmm - I'm not sure how helpful the `--debug` stuff is.  Shouldn't the statement about `JUPYTER_PREFER_ENV_PATH` be displayed unconditionally?  

I think making the print statement unconditional (relative to a debug flag) is probably the best we can do and hope they've set the env at a more global scope (such that `jupyter --paths` is running in the same env as the application).  I tend to lose sight of the fact that these are self-managed servers (for the most part) and the "administrator" should be aware of the env setting.  As such, I don't think introducing the `--debug` switch just for this feature is worthwhile.

jasongrout:
> I think making the print statement unconditional (relative to a debug flag) is probably the best we can do

> As such, I don't think introducing the `--debug` switch just for this feature is worthwhile.

Not sure what you'd like here. As one more effort, I made `--debug` report the status of each environment variable and how it is affecting the paths, in an effort to help people debug their issues. If this still doesn't seem helpful, we can revert adding the `--debug` switch.

jasongrout:
Here is what it looks like in practice (with home and environment level directories elided for privacy):

```
% JUPYTER_PREFER_ENV_PATH=y JUPYTER_PATH=a:b JUPYTER_CONFIG_DIR=user-config-dir jupyter --paths --debug
JUPYTER_PREFER_ENV_PATH is set, making the environment-level path preferred over the user-level path for data and config
JUPYTER_NO_CONFIG is not set, so we use the full path list for config
JUPYTER_CONFIG_PATH is not set, so we do not prepend anything to the default config paths
JUPYTER_CONFIG_DIR is set to 'user-config-dir', overriding the default user-level config directory
JUPYTER_PATH is set to 'a:b', which is prepended to the data paths
JUPYTER_DATA_DIR is not set, so we use the default user-level data directory
JUPYTER_RUNTIME_DIR is not set, so we use the default runtime directory

config:
    /[environment directory]/etc/jupyter
    user-config-dir
    /usr/local/etc/jupyter
    /etc/jupyter
data:
    a
    b
    /[environment directory]/share/jupyter
    /[home directory]/Library/Jupyter
    /usr/local/share/jupyter
    /usr/share/jupyter
runtime:
    /[home directory]/Library/Jupyter/runtime
```

kevin-bates:
Jason and I spoke offline to get in sync here and, with the latest changes that dump the status of supported envs, we're good moving forward here.  Thank you Jason!

kevin-bates (approved):
Thank you Jason.

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/jupyter-paths-priority-order/7771/1

PR #200, opened 2020-10-29
Fix python 3 requirements

jasongrout:
Even though python 3.5 is EOL now, it did just have its final release a month ago, and we do not actually have 3.6-specific code, so it seems nice to those still migrating off of 3.5 to let them have the other features the next release brings.


jasongrout:
If the next release of jupyter_core is a major release, then I fully support moving to python 3.6 as a requirement...

jasongrout:
Superseded by #201

PR #201, opened 2020-10-29
Add a GitHub Action test workflow

jasongrout:
It looks like travis is taking a *long* time these days to test PRs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,51 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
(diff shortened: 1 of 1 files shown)


jasongrout:
Python 3.5 was giving test errors on Windows. It's EOL now - I think it's a good time to drop support for it.

.github/workflows/test.yml
```diff
@@ -0,0 +1,46 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
+
+on:
+  push:
+    branches: [ master ]
```
kevin-bates, line 8: Could we open this up so that folks can ensure their potential PR builds prior to submission?
```suggestion
    branches: '*'
```

Many projects do the same for `pull_request.branches` below, which would be my preference, but, I believe, that can lead to double builds on pushed updates to existing PRs.  Since these builds are only 2-3 minutes, I think the additional builds aren't a big deal.

kevin-bates (commented):
Thanks Jason - this is a great update!  Just had the one suggestion/comment regarding the branches section.

PR #203, opened 2020-10-31
Revert the change to parsing in JUPYTER_NO_CONFIG in #199

jasongrout:
That was technically a breaking change, so should not have been done in a minor release.

We can make all environment variable boolean flags consistent in 5.0.

@kevin-bates - after thinking more about this, I think we should play it very safe and conservative here, and preserve behavior until 5.0.


kevin-bates (approved):
I'm good with waiting on this update until 5.0.  Thanks.

PR #204, opened 2020-10-31
Update changelog for 4.7

jasongrout:
I also updated the version info file to follow more pythonic conventions.

### DIFF
--- a/jupyter_core/version.py
+++ b/jupyter_core/version.py
@@ -1,6 +1,24 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-version_info = (4, 6, 4, "dev")
+from collections import namedtuple
 
-__version__ = '.'.join(map(str, version_info))
+VersionInfo = namedtuple('VersionInfo', [
+    'major',
+    'minor',
+    'micro',
+    'releaselevel',
+    'serial'
+])
+
+version_info = VersionInfo(4, 6, 4, 'alpha', 0)
+
+_specifier_ = {'alpha': 'a', 'beta': 'b', 'candidate': 'rc', 'final': ''}
+
+__version__ = '{}.{}.{}{}'.format(
+    version_info.major,
+    version_info.minor,
+    version_info.micro,
+    (''
+     if version_info.releaselevel == 'final'
+     else _specifier_[version_info.releaselevel] + str(version_info.serial)))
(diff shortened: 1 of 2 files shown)


jupyter_core/version.py
```diff
@@ -1,6 +1,24 @@
...
+    'major',
+    'minor',
+    'micro',
+    'releaselevel',
+    'serial'
+])
+
+version_info = VersionInfo(4, 6, 4, 'alpha', 0)
+
+_specifier_ = {'alpha': 'a', 'beta': 'b', 'candidate': 'rc', 'final': ''}
```
kevin-bates, line 16: Is dropping the developmental specifier intentional?  I view all these as being tied to an actual release or pre-release posted to pypi while `dev` is something I know is locally built.

I'd be in favor of adding a `'development': 'dev'` or `'dev': 'dev'` entry, but if that's not the convention then I suppose that's okay, although a number of Jupyter repos still use 'dev'.

kevin-bates (commented):
Thanks Jason. The changelog looks great.  Just had the one comment/question regarding the change to the "default" version info specifier.

jupyter_core/version.py
jasongrout: Sure. We were going off of the python conventions in the python source (https://github.com/python/cpython/blob/4b9aad49992a825d8c76e428ed1aca81dd3878b2/Python/sysmodule.c#L2534-L2541), but I see that pep 440 also has development releases as an official convention: https://www.python.org/dev/peps/pep-0440/#developmental-releases

I'll add a `'development': 'dev'` entry too.

kevin-bates: Ah, yeah, I figured you were using PEP 440.  Thanks.  Should the suffix be '.dev' in this case - since that's the current convention?

jasongrout: Yep, I fixed that too in the commit I just pushed.

jasongrout:
I'll push another PR that actually supports using dev according to pep440 (i.e., prereleases should be able to be listed as dev, for example). We can (a) decide if the extra complexity is worth it, and (b) merge it for the next major version, since you might consider moving version_info from `(4, 6, 4, 'dev')` to a pattern like `(4, 6, 4, 'final', 0, development=True, developmentserial=0)` to be a breaking change for people doing development.

blink1073 (approved):
Thanks!

PR #206, opened 2020-11-02
Make dev releases follow pep 440

jasongrout:
This follows on #204 (and it includes it, mostly) to make dev alpha/beta/rc releases possible, as pep440 indicates should be possible.

This also complicates the version info and file more. It also disrupts the existing convention that `version_info[3] == 'dev'` during a dev release, so it may not be considered backwards compatible. So we need to decide if this extra capability is worth it, and if so, should it be released in a major version.

Or we decide it isn't worth the extra complexity and close this pr.

### DIFF
--- a/jupyter_core/version.py
+++ b/jupyter_core/version.py
@@ -8,18 +8,21 @@ VersionInfo = namedtuple('VersionInfo', [
     'minor',
     'micro',
     'releaselevel',
-    'serial'
+    'serial',
+    'development',
+    'developmentserial'
 ])
 
-version_info = VersionInfo(4, 6, 4, 'dev', 0)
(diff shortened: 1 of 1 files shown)


blink1073:
This has a conflict now.

kevin-bates:
I personally found the previous change where the specifier contained the `'dev'` entry easier to follow.

jasongrout:
I fixed the conflict. Regardless, this probably shouldn't be merged before 5.0 even if we decide to merge it.

> I personally found the previous change where the specifier contained the `'dev'` entry easier to follow.

Me too, even if it didn't really follow the spirit of pep440.

I'll close this for now, but if we want it, we can always open it back up.

PR #209, opened 2020-11-21
data/config path entry_points with minimal examples

bollwyvl:
## Background

Jupyter relies on a hierarchy of directories (user-level, environment-level, system-level, etc.) to store configuration and data. These directories are used by a number of Jupyter programs, for example:
- Most applications based on the traitlets Configurable application class store configuration in JSON files in the configuration directories. They also aggregate conf.d-style configuration from these directories to determine settings of options.
- Jupyter Notebook extensions copy their javascript assets into a data directory on installation for the server to serve
- JupyterLab extensions copy their javascript assets into a data directory on installation for the server to serve.

## Problem

Currently the environment level of this directory hierarchy is a fixed location based on `sys.prefix`. This means that packages need to copy their files into this directory at install time, which has several issues:

- Copying files into a data directory uses the `data_files` feature of Python packages, which is [deprecated](https://setuptools.readthedocs.io/en/latest/references/keywords.html?highlight=data_files) in setuptools and is not supported in non-setuptools-based packagers like `flit`, `poetry` (see [here](https://github.com/python-poetry/poetry/issues/890)), etc.
- Data files are duplicated in the package bundle (once for copying into the data directory, once for being included in the actual package to install into `site-packages`). For some extensions, this a huge (like megabytes or [tens of megabytes](https://pypi.org/project/ipydrawio/#files)).
- Development installs (`pip -e`) do not update data files when the source files change, so when developing a package, if something changes to the data files, you either have to copy them over again, or you have to run a command to make the appropriate data directory a symbolic link (not available on some platforms) to the source files.

(Also, it seems that sometimes these data file directories are not deleted. For example, in JupyterLab we actually create files at runtime in the data directory, and I think they don't get deleted when JupyterLab is uninstalled)

## Proposed solution

Python has another mechanism that is explicitly designed for plugin systems called *entry points*. An entry point is a piece of metadata in a package that points to an arbitrary import from the package. This PR changes `jupyter_core` to look for two specific entry points in any installed package, each pointing to a list of paths, to augment the environment-level Jupyter config directories (the `jupyter_config_paths` entry point) and data directories (the `jupyter_data_paths` entry point). The result is:

- Any package can add new environment-level Jupyter config and data directories. In practice, this means that a package can contain data or configuration in a directory that is installed in its `site-packages` directory, and can use the entry point to point Jupyter to that internal directory. Since this directory is internal to the package:
  - the files are not duplicated in the package tarball
  - development (un)installs automatically work, since the directory points to an internal directory in the package
  - other python package managers can be used, like poetry using its include/exclude mechanism for files
- non-Python programs can access this (and all other paths) by shelling out to `jupyter --paths --json`

### Problems with the proposed solution

- Entry points are based on importing a module to get a value, which potentially could be very expensive. We [explore](https://github.com/jupyter/jupyter_core/pull/209#issuecomment-792049406) parsing the file first for literal values, and then importing as a last resort, which seems to alleviate this problem in the common case (setuptools does something similar for its `attr` handler for setup.cfg values).
- neither `entry_point` group is cached
  - an interactive installation with e.g. `pip install` or `conda install` would be able to update the search path, provided the application isn't doing its own caching...
    - this is important to maintain the observed behavior of `data_files`
    - because the import system is invoked, users of this system may wish to create a separate `python_packages` entry for these static assets, to avoid bringing in otherwise-unused runtime dependencies, e.g. `pandas`
  - [ ] adding some debug logging around this will help pinpoint slow startup times
    - turns out there is no logging this deep in the stack. we could either:
      - [ ] add a `log=None` argument to the various calls
      - [ ] add a logger controlled by a `JUPYTER_CORE_LOGLEVEL`
- if an `entry_point` is _added_ or (its target is _changed_) in a package with an editable install, it must be reinstalled
  - however, if only the _return value_ of an _existing_ `entry_point` is changed, no re-install is required
- existing tools that were relying on indexing  `jupyter_*paths()`
  - this occurs in the test suite for `jupyter_core` itself: if one of the example packages is installed, the tests break
  - [ ] these will have to be updated to inspect relative positions, e.g. _was the user dir loaded before or after the env paths when `JUPYTER_PREFER_ENV` is set_

### Alternative solutions

setuptools also provides a way for a package to have custom metadata files in the egg or dist_info directories. This avoids the problems of importing or parsing an arbitrary python file to get the few strings that we need. However, it appears that this arbitrary metadata is not well supported outside of setuptools. See [below](https://github.com/jupyter/jupyter_core/pull/209#issuecomment-791400757) for some experiments around this approach.

## Example

See the [setuptools example](https://github.com/jupyter/jupyter_core/tree/38e3acd220153871ddd93d3e77a8b0af9e18c9db/examples/jupyter_path_entrypoint_setuptools), specifically https://github.com/jupyter/jupyter_core/blob/38e3acd220153871ddd93d3e77a8b0af9e18c9db/examples/jupyter_path_entrypoint_setuptools/setup.cfg#L35-L39

- this approach requires a boilerplate `MANIFEST.in` and a `setup.py` in order to be installed from source

and the [flit example](https://github.com/jupyter/jupyter_core/tree/38e3acd220153871ddd93d3e77a8b0af9e18c9db/examples/jupyter_path_entrypoint_flit), specifically https://github.com/jupyter/jupyter_core/blob/38e3acd220153871ddd93d3e77a8b0af9e18c9db/examples/jupyter_path_entrypoint_flit/pyproject.toml#L11-L15
for examples of how to use these entry points.

- `pyproject.toml` is the only boilerplate file needed, and _generates_ a `setup.py`
- `flit` can also generate binary reproducible `whl` files (for python >=3.7) given the same version of `flit_core`

## Original issue description

Hey folks! Thanks for keeping this foundational technology working.

`data_files` are making me sad enough that I'm willing to [bring this up again](https://github.com/jupyter-server/jupyter_server/issues/351).

This is a low-downstream-impact way we could allow python packages to not require the ill-supported `data_files` technique. 

To test:

```bash
pip install -e .
cd examples/entry_point_example
pip install -e .
jupyter --paths
# should see that development environment in place
pip uninstall entry_point_example
jupyter --paths
# it's gone
```

I don't know if it _really_ works yet, down the the _n_-th downstream, but seems it _should_ if they are relying on `jupyter_*_dir`, and handling multiple paths already.


maartenbreddels:
I see this as a good alternative to using data_files without overhauling the config system. I am a bit worried that it's hard to debug when things go wrong (if 15 directories will be scanned). Could we maybe provide a richer debug facility to see a particular config key, and how each directory is changing it. Grepping in 15 directories will not be fun. Or do I see a problem that does not exist, and are the debug options sufficient?

bollwyvl:
>  Grepping in 15 directories will not be fun

Yep, there will be a lot of directories beyond the Big Four. No doubt some combination of `jupyter --paths`, `jq`,  and `xargs` would make `grep` plausible, but that's no fun!

A `JupyterApp` base flag like `--show-config` which every app would inherit is a whacking good idea, even outside of this little draft. It could probably use `difflib` to generate a decently-readable representation of the config before each file was loaded, and show the final config, perhaps something like:

```bash
$> jupyter foo --show-config

environment variables:
- JUPYTER_PREFER_ENV_PATH: not set
- ...

paths:
- /etc/jupyter/jupyter_config.json: not found
...
- ~/my-project/src/my_project/etc/jupyter_foo_config.d/my-project.json:

    + SomeHasTraits:
    +   foo: bar
...
- ~/my-project/src/my_project/.venv/etc/jupyter_config.d/someone-elses-project.json:

      SomeHasTraits:
    -   foo: bar
    +   foo: baz

...
- ./jupyter_foo_config.json: not found

final:

    SomeHasTraits:
      foo: baz
```

sprinkle in some pygments (if available) and it would be pretty usable.

maartenbreddels:
Indeed, exactly what I had in mind, that would help a lot

bollwyvl:
Gah, looking at it: a lot of the complexity is duplicated between `jupyter_server` and `notebook`... while both would work with this PR, there's no simple way to add the above config inspection.

Perhaps the better short-term approach would be to invert it, with a separate package/command, e.g. offered `jupyter show-config notebook FooHasTraits.bar`. I guess this would work by overloading/monkeypatching `config_manager_class` (gaaah) with an instrumented subclass, and call `initialize` but not `start`.

Because of that complexity, this could probably _not_ land here, unless the ConfigManager pattern was brought upstream, which sounds hard to coordinate.

bollwyvl:
I have an unshaeably bad version of this, but it kinda works with `notebook`, `jupyter_server`, `jupyterlab` and `voila` installed:
```
getting jupyter_server_config from /etc/jupyter
got {}
getting jupyter_server_config from /usr/local/etc/jupyter
got {}
getting jupyter_server_config from /home/weg/projects/jupyter_showconfig_/envs/default/etc/jupyter
Reading file /home/weg/projects/jupyter_showconfig_/envs/default/etc/jupyter/jupyter_server_config.d/jupyterlab.json
Reading file /home/weg/projects/jupyter_showconfig_/envs/default/etc/jupyter/jupyter_server_config.d/nbclassic.json
Reading file /home/weg/projects/jupyter_showconfig_/envs/default/etc/jupyter/jupyter_server_config.d/voila.json
got {'ServerApp': {'jpserver_extensions': {'jupyterlab': True, 'nbclassic': True, 'voila.server_extension': True}}}
getting jupyter_server_config from /home/weg/.jupyter
got {}
getting page_config from /etc/jupyter/labconfig
got {}
getting page_config from /usr/local/etc/jupyter/labconfig
got {}
getting page_config from /home/weg/projects/jupyter_showconfig_/envs/default/etc/jupyter/labconfig
got {}
getting page_config from /home/weg/.jupyter/labconfig
got {}
[I 2020-11-22 17:50:37.177 ServerApp] jupyterlab | extension was successfully linked.
getting jupyter_notebook_config from /home/weg/.jupyter
got {}
getting jupyter_notebook_config from /etc/jupyter
got {}
getting jupyter_notebook_config from /usr/local/etc/jupyter
got {}
getting jupyter_notebook_config from /home/weg/projects/jupyter_showconfig_/envs/default/etc/jupyter
Reading file /home/weg/projects/jupyter_showconfig_/envs/default/etc/jupyter/jupyter_notebook_config.d/jupyterlab.json
Reading file /home/weg/projects/jupyter_showconfig_/envs/default/etc/jupyter/jupyter_notebook_config.d/voila.json
got {'NotebookApp': {'nbserver_extensions': {'jupyterlab': True, 'voila.server_extension': True}}}
getting jupyter_notebook_config from /home/weg/.jupyter
got {}
[I 2020-11-22 17:50:37.322 ServerApp] nbclassic | extension was successfully linked.
[I 2020-11-22 17:50:37.322 ServerApp] voila.server_extension | extension was successfully linked.
[I 2020-11-22 17:50:37.339 LabApp] JupyterLab extension loaded from /home/weg/projects/jupyter_showconfig_/envs/default/lib/python3.7/site-packages/jupyterlab
[I 2020-11-22 17:50:37.339 LabApp] JupyterLab application directory is /home/weg/projects/jupyter_showconfig_/envs/default/share/jupyter/lab
[I 2020-11-22 17:50:37.342 ServerApp] jupyterlab | extension was successfully loaded.
[I 2020-11-22 17:50:37.345 ServerApp] nbclassic | extension was successfully loaded.
[I 2020-11-22 17:50:37.347 ServerApp] voila.server_extension | extension was successfully loaded.
```

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/how-do-we-uninstall-extensions-that-have-been-installed-using-jupyter-labextension-develop-overwrite/7845/5

jasongrout:
@bollwyvl - I made a PR to your PR with a few changes I thought would be good: https://github.com/bollwyvl/jupyter_core/pull/1. What do you think?

bollwyvl:
@jasongrout thanks for that! I merged in upstream, added a flit example and some tests.

jasongrout:
Overall, I think this is a good idea. My immediate usecase is using this for installing prebuilt extensions, and originally I was thinking that we just enable an entry point for jlab plugins specifically. I like that this solution is far more general and more foundational than a jlab-specific entry point!

bollwyvl:
Yeah, federated extensions are a big motivator: as mentioned, the `data_files` approach yields duplicates in wheels, which really starts adding up on large builds (wasm, design templates, etc).

I really like how dense the `flit` example is... there's _sane_ globbing/excluding, and the reproducible wheel building is a big step forward. I would move all of my labextensions to that toolchain as soon as this was available.

jasongrout:
@bollwyvl - do you mind if I edit the original description to provide an overview of this PR and its impact on how things work in Jupyter?

bollwyvl:
Free copy editing!? This place has everything!

blink1073 (approved):
Amazing!

bollwyvl:
Some totally unscientific timings with the two tiny example packages:

From the CLI:
```
Init jupyter_config_paths in 3.5849ms
Init jupyter_config_paths:entry-point-example-setuptools in 0.1771ms
Init jupyter_config_paths:entry-point-example-flit in 0.1349ms
Init jupyter_data_paths in 3.1528ms
Init jupyter_data_paths:entry-point-example-setuptools in 0.0212ms
Init jupyter_data_paths:entry-point-example-flit in 0.0160ms
```

From an interactive `python` session (not ipython):
```
Init jupyter_data_paths in 8.7347ms
Init jupyter_data_paths:entry-point-example-setuptools in 0.2825ms
Init jupyter_data_paths:entry-point-example-flit in 0.1893ms
Init jupyter_config_paths in 5.3611ms
Init jupyter_config_paths:entry-point-example-setuptools in 0.0370ms
Init jupyter_config_paths:entry-point-example-flit in 0.0262ms
```

I'd wager the baseline init of finding all the entry_points is static, and will increase linearly with the number of entry_points loaded. Subsequent calls take the same amount of time. 

If, as is likely, the two entry point targets are both loaded from `__init__.py`, the load import cost will only be paid the first time, as everything will be cached in the import machinery.

However, if each `GET` request to a Jupyter server requires warming these up, this will add up rapidly. So we'd need to decide where it would be appropriate to control caching, such as (not mutually exclusive):
- a sensible, environment-variable override-able default cache period (e.g. a minute)
- offer a `force=False` flag for cache invalidation
  - e.g. an app's `index.html`-equivalent handler could invalidate the cache when it starts a template render, but subsequent requests for images/js/css would not

jasongrout:
> However, if each `GET` request to a Jupyter server requires warming these up, this will add up rapidly. So we'd need to decide where it would be appropriate to control caching, such as (not mutually exclusive):

For loading JLab extensions (IIRC), we cache the information once at application startup.

bollwyvl:
Welp, if you're on binder, and "pip install" a widget library, you'd
refresh the page... And not get your widgets. But your kernel-side stuff
would work.

I ran into a similar issue in the licenses pr (also the federated extension
info _appears_ to be in a closure).

jasongrout:
Here's another crazy idea that works around the problem of entry points always importing a package's __init__.py, which may be expensive:

Using https://setuptools.readthedocs.io/en/latest/userguide/extension.html#adding-new-egg-info-files as a basis, instead of using entry points, I use custom package metadata. My changes are in this commit: https://github.com/jasongrout/jupyter_core/commit/66351b0978f21b275dd58ab13e6448ef3c610705. Since I'm poking around in setuptools, which I don't know much about, I'm sure the code could be cleaned up or made more general by someone more familiar with setuptools.

The tradeoff is that the paths are more declarative, and I assume that paths are relative to the package root, rather than being able to compute them on the fly like with entry points.

bollwyvl:
> custom package metadata.

My issue is: I think if we're going to do something _special_ for python packages (as opposed to julia, r, etc. to which this information will be _pretty much_ opaque), my preference would be to stay as close to the spec (PEP [517](https://www.python.org/dev/peps/pep-0517/), [639](https://www.python.org/dev/peps/pep-0639/), ...) as possible, in a format that as many tools as possible support.  Introducing new metadata fields, that can _only_ be feasibly created with `jupyter_packaging`, will disincentivize maintainers creating packages that "casually" work with Jupyter tools.

I think with sound documentation (that _could_ be under `jupyter_packaging`):
- caching
- providing profiling information through some means
- guidance on making the packages that own these _dedicated_ modules (as in, shows up in `top_level.txt`), with (almost no imports) 

...`entry_points` is a fine choice, and are already used/documented in `nbconvert` and elsewhere. 

And, back to the lab extensions point: I _might just_ be thinking about some ability to dynamically generate federated extensions without nodejs, much less `jupyter_packaging`, and making this `entry_point` target a singleton with `__slice__` method...

bollwyvl:
> custom package metadata.

But thank you @jasongrout for providing additional options. Please feel free to hoist that idea up to the description! I took `[strawman]` off, but the intent of the original issue (https://github.com/jupyter-server/jupyter_server/issues/351) still stands... I'm always sad when a snap decision gets made just because _someone needs it/is getting paid for it today_ or whatever, rather than _we think this is good for all Jupyter stakeholders, from the janitors to the hot-shot spaceship pilots_.

jasongrout:
Are these two statements referring to the same concept?

>  because the import system is invoked, users of this system may wish to create a separate `python_packages` entry for these static assets, to avoid bringing in otherwise-unused runtime dependencies, e.g. `pandas`

> guidance on making the packages that own these _dedicated_ modules (as in, shows up in `top_level.txt`), with (almost no imports)

If so, can you elaborate? Is there a way to have a package that may have a relatively expensive top-level import, but still have a relatively lightweight entry point by having the entry point not requiring the expensive import?

jasongrout:
>  Introducing new metadata fields, that can _only_ be feasibly created with `jupyter_packaging`, will disincentivize maintainers creating packages that "casually" work with Jupyter tools.

Note that the custom metadata here only relies on jupyter_core as a dependency, not jupyter_packaging. The example was updated to show that all a user would have to do is add one or two arguments to their setup.py file (or equivalent entries in setup.cfg), and have a setup_requires point to jupyter_core.

jasongrout:
> If so, can you elaborate? Is there a way to have a package that may have a relatively expensive top-level import, but still have a relatively lightweight entry point by having the entry point not requiring the expensive import?

For completeness: I also explored a little bit using namespace packages to try to avoid importing the main package to get to a lightweight entry point, but gave up as being too complicated/magical to recommend to everyone.

jasongrout:
> my preference would be to stay as close to the spec (PEP [517](https://www.python.org/dev/peps/pep-0517/), [639](https://www.python.org/dev/peps/pep-0639/), ...) as possible, in a format that as many tools as possible support.

If other packaging tools do not support custom metadata, then that would be a showstopper for using custom metadata, I think.

I agree the entry points would be better theoretically, it just bothers me that you have to import the package to get at this static configuration data.

jasongrout:
> but a distribution can....
> then `find_packages(include="src")` would turn up `my_widget` and `my_widget_config`

In this case, would we have:
* A single tarball/wheel distributed via pypi, or multiple tarball/wheel files?
* two top-level imports in python, i.e., `import my_widget` and `import my_widget_config` both would work?
* two top-level directories in site-packages, or one?

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/jupyter-paths-priority-order/7771/3

bollwyvl:
in the simplest case:
- a single `tar.gz`, a single `.whl`
- two top-level imports
- two top-level directories in `site-packages`

again, when folk use `find_packages` today, this happens _all the time_ with a `site-packages/tests`. Even `jupyter_server` did this for a while! But we can use the heck out of it. These things won't be `zip_safe`... but they never were.

ancedotally, on jupyterlab-lsp, we opted to split them, and `jupyterlab-lsp` has a dependency on `jupyter-lsp`, as it would work with any `jupyter_server` and not bring in jlab-specific deps like `json5`.

jasongrout:
> this happens all the time with a site-packages/tests

Very interesting! Does pip not give any warning when a package installation overwrites an existing directory, i.e., two different packages stomp on each other?

jasongrout:
If other packaging systems support custom metadata, I think that's probably cleaner than having a separate top-level directory and top-level import just for an entry point, i.e., you only have one top-level import/directory. On the other hand, maybe that extra package is only done in situations where the original package import is expensive, and maybe that's not so common. On the other hand, maybe that second package contains the actual etc/ and share/ directories as well (e.g., that second package is what distributes the javascript assets for a lab extension) - perhaps it is a bit iffy to assume you can do `../mypackage/etc/jupyter` to point to the first package from the second package?

bollwyvl:
> If other packaging systems support custom metadata

I haven't looked into it... i loathe making PRs to packaging systems to support stuff. I mean, the ground is littered with old `Please make AMD module` PRs to have stuff work with `requirejs`, `plz publish language server on npm` for LSP... but at least these made things better for _all_ their downstreams. If custom metadata is not wide-spread (and/or doesn't has a _formal_ PEP), i just dunno... it sure sounds like rewriting `entry_points`, But For Jupyter... which is my point in this whole exercise.

>  one top-level import/directory. 

Welp... most folk don't want to tell people to `import my_widget.widgets_foo`. You can do magic `__getattr__` and stuff to make them lazy, though. But if _that_ `widgets_foo.py` imports `pandas`, then were back at the initial expensive load.

> Does pip not give any warning

Nope, otherwise old-style namespace packages like `backports.*` wouldn't work, but `conda`, `apt`, etc. sure do, and have to patch around these things! `pip` also ignores `extras` pins after the fact, yadda, yadda... it's not great, but it's what we have to deal with. We do _not_ what, A Package Manager, But For Jupyter (though I'll stand by `jlpm` being a good call :blush:).

> second package contains the actual etc/ and share/ directories as well 

Yeah, that pattern that could work... certainly for `etc`. None of the `_jupyter_this_and_that_extension` magic functions need any imports. 

But then, many of those `enable --py`, etc. won't even be needed for something that uses these tools, just `pip install -e` (or `flint install --symlink`). 

With federated extensions, folks _are_ reading more JSON than before, which is good for DRY, and that isn't _free_, but usually that's a one-time hit. 

> bit iffy to assume you can do ../mypackage/etc/jupyter to point to the first package from the second package?

I mean, that's what you buy when you publish packages that end up in `site-packages`. But sure. Especially with `--symlink`, stuff gets weird.

jasongrout:
>  i loathe making PRs to packaging systems to support stuff.

Yep, agreed. If other packaging systems don't support the custom metadata, I think the option is out.

bollwyvl:
> bit iffy to assume you can do ../mypackage/etc/jupyter to point to the first package from the second package?

It's also possible to find a [module spec](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec) in a distribution, and then grab it's `spec.__file__` _without_ importing it, from which the location of `share` can be derived, e.g. `Path(spec.__file__).parent / "share"`. 

This mechanism is probably _not_ appropriate for the overall mechanism (even if an `entry_point` does expose its path, which i haven't looked into), as having at least _some_ dynamism/irreproducibility is more useful than requiring the files be in a specifically-named file in that path.

bollwyvl:
A little bird told me about the pattern in [StaticModule](https://github.com/pypa/setuptools/blob/23ee037d56a6d8ab957882e1a041f67924ae04da/setuptools/config.py#L19).

If we _did_ use this, the documented contract could be something like:

> The `entry_point`'s `module.submodule:importable` MUST be a literal list or tuple of POSIX paths (e.g. with `/` not `\`) relative to the file of `module.submodule`.

This would improve the stability of the system, as no code would be executed, as the actual target module wouldn't be loaded, and the performance, as we could rely on the import caching mechanisms, which are already quite robust. This leaves the overhead of enumerating each `entry_point` group, but that can still be cached for a reasonable amount of time, and is _at least_ going to be predictable, modulo the number of packages installed.

jasongrout:
> A little bird told me about the pattern in [StaticModule](https://github.com/pypa/setuptools/blob/23ee037d56a6d8ab957882e1a041f67924ae04da/setuptools/config.py#L19).

I had wondered about how setup.py implemented this note in its documentation for setup.cfg [attr](https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#specifying-values):

> In order to support the common case of a literal value assigned to a variable in a module containing (directly or indirectly) third-party imports, attr: first tries to read the value from the module by examining the module’s AST. If that fails, attr: falls back to importing the module.

My guess is that this is how it is doing it. You're right - perhaps that is the right middle ground. Or perhaps we just do what attr does and first try to read it statically, and if that doesn't work, *then* try the import. Then we have fast reading the vast majority of the time, but can fall back to the full entry point load if the dynamic entry point is really needed.

bollwyvl:
Ooh, cake and eating, too!

jasongrout:
Huh, I'm curious for simple cases if ast-parsing the module is faster or slower than actually executing the precompiled .pyc file. Are we actually going to slow things down a lot for the common case by parsing the source code, compared to executing the already-parsed (I think?) .pyc file?

bollwyvl:
I think both are fed by the line cache, but there's not way to tell without
some profiling.

But it's not the parsing, it's the chance for knock-on imports and side
effects that would most likely introduce non-determinism.

jasongrout:
> it's the chance for knock-on imports and side effects that would most likely introduce non-determinism.

Agreed. I think it's still worth pursuing this parsing, then loading strategy. I was just curious

bollwyvl:
Took a sec, and some cargo-culting, but now have that in.

yeah, parsing the ast is, on the main, slower than importing (of my tiny examples).

```
JUPYTER_ENTRY_POINT_STRATEGY="PARSE" JUPYTER_ENTRY_POINT_TIMINGS=1 apr time jupyter --paths
3.65ms jupyter_config_paths loaded
0.35ms	jupyter_config_paths	entry-point-example-setuptools
0.25ms	jupyter_config_paths	entry-point-example-flit
4.28ms jupyter_config_paths	TOTAL
3.28ms jupyter_data_paths loaded
0.11ms	jupyter_data_paths	entry-point-example-setuptools
0.11ms	jupyter_data_paths	entry-point-example-flit
3.52ms jupyter_data_paths	TOTAL
```

```
JUPYTER_ENTRY_POINT_STRATEGY="LOAD" JUPYTER_ENTRY_POINT_TIMINGS=1 apr time jupyter --paths
3.69ms jupyter_config_paths loaded
0.26ms	jupyter_config_paths	entry-point-example-setuptools
0.24ms	jupyter_config_paths	entry-point-example-flit
4.22ms jupyter_config_paths	TOTAL
3.25ms jupyter_data_paths loaded
0.10ms	jupyter_data_paths	entry-point-example-setuptools
0.11ms	jupyter_data_paths	entry-point-example-flit
3.49ms jupyter_data_paths	TOTAL

```

bollwyvl:
If, however, one of those triggers an expensive import (in this case, `pandas`):

```
JUPYTER_ENTRY_POINT_STRATEGY="PARSE" JUPYTER_ENTRY_POINT_TIMINGS=1 apr time jupyter --paths
4.16ms jupyter_config_paths loaded
0.39ms	jupyter_config_paths	entry-point-example-setuptools
0.26ms	jupyter_config_paths	entry-point-example-flit
4.86ms jupyter_config_paths	TOTAL
3.65ms jupyter_data_paths loaded
0.08ms	jupyter_data_paths	entry-point-example-setuptools
0.09ms	jupyter_data_paths	entry-point-example-flit
3.85ms jupyter_data_paths	TOTAL

```

```
JUPYTER_ENTRY_POINT_STRATEGY="LOAD" JUPYTER_ENTRY_POINT_TIMINGS=1 apr time jupyter --paths
4.21ms jupyter_config_paths loaded
404.67ms	jupyter_config_paths	entry-point-example-setuptools   # woof
0.27ms	jupyter_config_paths	entry-point-example-flit
409.19ms jupyter_config_paths	TOTAL
3.96ms jupyter_data_paths loaded
0.11ms	jupyter_data_paths	entry-point-example-setuptools
0.12ms	jupyter_data_paths	entry-point-example-flit
4.22ms jupyter_data_paths	TOTAL

```

bollwyvl:
I also haven't figured out how we might (ab)use the loader cache, which might be smarter about things than `open().read()`... poring over [importnb](https://github.com/deathbeds/importnb/) at present...

jasongrout:
Looking at poetry, I see at least some place it handles [dist info files](https://github.com/python-poetry/poetry/blob/adc3cde3473e85eb744f1f73d937439a681ae365/poetry/masonry/builders/editable.py#L196-L248), and it doesn't seem to have a provision in there for custom metadata files in the dist_info directory.

jasongrout:
As a check on our assumptions and conclusions about python packaging, I've asked @gaborbernat to comment.

@gaborbernat - here's a short summary (a much more detailed summary is in the original description):

Jupyter currently has an ecosystem where individual packages can contribute resources or configuration (kind of like a plugin ecosystem for Jupyter). Currently each python package that wants to contribute resources uses the setuptools `data_files` functionality to put these files in a central location under sys.prefix. However, there are a number of problems with that, including `data_files` being deprecated by setuptools and not being supported by other packagers like flit or poetry.

In this PR, we've explored using entry points to expose these resources to Jupyter. One of the concerns with using entry points is that querying them requires importing python modules, and that is potentially very expensive. We've also explored doing something similar to setuptools where we parse the entry point module instead of importing it, which seems to cut down on the cost. We've also explored using setuptools to [add new metadata files](https://github.com/jupyter/jupyter_core/pull/209#issuecomment-791400757) directly, to avoid this importing issue, but that seems like it is not supported at all by other packagers too. I also explored somehow encoding this data into the PKG_INFO files as package metadata, but that seems like a closed standard that does not really support arbitrary metadata additions.

As it stands now, it looks like the best option is to expose an entry point, which is a list of directories relative to the entry point module, where Jupyter can look for resources. We would try parsing the entry point module first, and if that failed (i.e., if the entry point was not a literal value), we would load the entry point to get this list of directories.

Do you have any suggestions for us that we might have missed for having python packages make metadata values like a list of directories available to (python) Jupyter programs, or more generally making arbitrary files available to Jupyter?

jasongrout:
@bollwyvl - perhaps we should reconsider (or consider separately?) entry points that are more specific than just adding directories to the Jupyter directory hierarchy. Perhaps it would be better to have a number of different entry points, one per type of information. For example, we could have a specific entry point for JupyterLab plugins, another specifically for traitlets config, another for notebook extensions. Then a python package that only provides a notebook extension will not be queried by Lab for lab extensions, and a python package providing a lab extension will not be queried for a notebook extension.

In other words, right now we distinguish between various pieces of information things want by subdirectories in our config and data directories. Perhaps with entry points we push that down one more level and we have a different entry point for each thing. I think it's equivalent information, but has some nice advantages:

1. different entry points could point to different files in a package's directory structure, rather than having to consolidate all of those things into a directory structure that mirrors Jupyter's directories. In other words, the notebook extension could be in `mypackage/nbextension`, and the labextension could be in `mypackage/labextension`, instead of having the package reconstruct a `data_dir/nbextension` and `data_dir/labextension` directory. In general, it's nice to be able to name things in a package according to what makes sense in the package (e.g., `labextension`) rather than having to conform to the directory structure that Jupyter insists on.
2. It might be nice to have multiple different entry points point to the same files. For example, often an ipywidgets JLab extension will work across multiple major versions of jlab. We could have an entry point for jlab 3 extensions, and another entry point for jlab 4 extensions, and a widgets extension could point both entry points to the same resource, where another package might point the jlab 3 entry point to a different bundle that the jlab 4 entry point.
3. For traitlets config, perhaps the entry point directly points to some JSON structure in python rather than a file on disk that contains JSON?
4. Lab would only look at the jlab-specific entry points, so would not have to parse/load packages that provided different jupyter entry points.

Some disadvantages: 
- there is a wider gulf between the Jupyter data/config directories and what entry points would give. We have to keep supporting the data/config directories for non-python solutions and for backwards compatibility, so we'd have to put up with two ways to get values.
- Each tool would have to somehow contain (or import from core?) the entry point processing code. The current approach having entry points for adding directories to the path Just Works if you use jupyter_core to get the directory hierarchy.
- It would be more confusing where values are coming from, and harder to debug unless/until each tool implemented some sort of debug mode that showed what entry points it was using and where it was inserting them into the hierarchy.

bollwyvl:
Or, going the other way: just a single `jupyter_path`, and you better darned know what `etc` and `share` look like. `pytest` has _1_ (it is _versioned_ though, and we may wish to consider that). nbconvert has _2_ ( `nbconvert.exporter` and `nbconvert.exporter.script`). 

For very deep things, this could become problematic, _especially_ on windows.  Federated lab extensions, in particular, which embed their full npm name, `%PREFIX%/Library/site-packages/my_widget_package/share/jupyter/labextensions/@my-organization/my-package/remoteEntry.<bigoldhash>.js` would be a _mess_ unless they're running right off `C:/`. But the existing approach isn't _too_ much different, i suppose, and practically would just save the `jupyter` bit.

My goal is not to supplant what already exists, and _works fine_ for most cases, or privilege python in any more than we must. `data_files` would be great, but I don't have the patience to make a PEP and then wait for a couple years. I just want to allow python packages to avoid `data_files`, and touch as little code as possible, and require as little boilerplate as possible, as that will lead to as few docs changes across _n_ sites, as possible, which will lead to as little ambiguity as possible, which will hopefully lead to as few issues as possible.

> sort of debug mode 

Yeah, I kinda wrote that... not sure where i put it. It basically would hook real deep into `traitlets.config.Application` (and the two JSONConfigManagers from `jupyter_server` and `notebook`) and then emitted stuff when those things changed. That would be useful _today_, but making it look nice isn't worth bringing in `rich` (or rewriting it by hand). Ideally it, and `conf.d` loading, would live in traitlets itself... but I'd rather see that work with `mypy` first :laughing: !

jasongrout:
> My goal is not to supplant what already exists, and _works fine_ for most cases, or privilege python in any more than we must. `data_files` would be great, but I don't have the patience to make a PEP and then wait for a couple years. I just want to allow python packages to avoid `data_files`, and touch as little code as possible, and require as little boilerplate as possible, as that will lead to as few docs changes across _n_ sites, as possible, which will lead to as little ambiguity as possible, which will hopefully lead to as few issues as possible.

Hmm, I think you're right. On with the current proposal, then!

jupyter_core/paths.py
```diff
@@ -9,20 +9,137 @@
...
+        vars(self).update(locals())
+        del self.self
+        # add a path
+        self.__file__ = spec.origin
+
+    def __getattr__(self, attr):
+        try:
+            return next(
+                ast.literal_eval(statement.value)
+                for statement in self.module.body
```
jasongrout, line 67: It seems like it should be going after the *last* literal assignment to mimic loading, where the last literal assignment wins over previous ones:

```suggestion
                for statement in reversed(self.module.body)
```

jasongrout: I filed an issue in setuptools about this at https://github.com/pypa/setuptools/issues/2602

bollwyvl: Sure, that makes sense... though when I've abused this for `__version__` in `setup.cfg`, it's always been one of those, "this is a singleton, mess around and find out" things. 

Would want a test, but not in the "simple" examples, probably...

jupyter_core/paths.py
```diff
@@ -9,20 +9,137 @@
...
+import pathlib
 import tempfile
 import warnings
+import functools
+import traceback
+import importlib
 
 from contextlib import contextmanager
 
+import entrypoints
```
gaborbernat, line 26: You should be using importlib.metadata instead here.

gaborbernat:
> As a check on our assumptions and conclusions about python packaging, I've asked @gaborbernat to comment.

In general here the idea is to use `importlib.metadata` and `importlib.resources`. Both of these are part of the standard library in newer releases, but are also available as backports (the backports mean you can use it today and don't have to wait until the standard library rolls out). See:

- https://importlib-resources.readthedocs.io/en/latest
- https://importlib-metadata.readthedocs.io/en/latest/using.html#entry-points

Something like:

```python
from contextlib import ExitStack
from importlib.metadata import entry_points
from importlib.resources import path

stack = ExitStack()
paths = set()
for group, eps in entry_points().items():
    if group == "x.y.z":
        for ep in eps:
            path = stack.enter_context(path(ep.module, ep.value))
            paths.add(path)
```

Note the above does not work yet, but I don't see a reason why we could not add support for it. Invoking the maintainers of those packages to clarify through @jaraco and @warsaw. They'll be able to give the definitive answer. I'd expect to be able to discover files without needing to import the package.

gaborbernat:
> my 2 cents: I know that `entrypoints` is orders of magnitude faster than `pkg_resources`, I'm not sure how importlib compares to that.

```entrypoints`` was the testing ground for ``importlib.metadata``. Nowadays the former is deprecated and replaced by the latter.

**Comment by jasongrout** [comment]
> Note the above does not work yet, but I don't see a reason why we could not add support for it.

Thanks @gaborbernat!

Right now this PR proposes an entry point like the following:

```
[options.entry_points]
jupyter_config_paths =
    entry-point-example-setuptools = entry_point_example_setuptools:JUPYTER_CONFIG_PATHS
```
where entry_point_example_setuptools.JUPYTER_CONFIG_PATHS is a list of directories. If I understand you correctly, you're suggesting that instead our entrypoint should point directly to the directory we'd like to add to our path hierarchy, like:
```
[options.entry_points]
jupyter_config_paths =
    entry-point-example-setuptools = entry_point_example_setuptools.etc.jupyter
```
(note that `entry_point_example_setuptools.etc.jupyter` is *not* an importable module, but it is the directory `entry_point_example_setuptools/etc/jupyter`). Then we'd retrieve the absolute path to that directory with importlib_resources.path. Am I understanding you correctly? Can we have an entry point that is not an importable module, and can we ask for a resource that is a directory?

**Comment by bollwyvl** [comment]
Some timings with replacing the `entry_points.get_group_named` with `importlib_metadata.entry_points` (haven't fixed the tests yet to work with it, much mock needed). Both of the below still use the parse-then-import approach, will take a look at that next (and probably add a wrapper that looks across all of these, runs a couple samples, etc.)

```
JUPYTER_ENTRY_POINT_TIMINGS=1 JUPYTER_ENTRY_POINT_FINDER=importlib_metadata jupyter --paths
26.95ms jupyter_config_paths loaded with importlib_metadata
0.30ms	jupyter_config_paths	entry-point-example-setuptools
0.25ms	jupyter_config_paths	entry-point-example-flit
27.54ms jupyter_config_paths	TOTAL
18.98ms jupyter_data_paths loaded with importlib_metadata
0.13ms	jupyter_data_paths	entry-point-example-setuptools
0.12ms	jupyter_data_paths	entry-point-example-flit
19.28ms jupyter_data_paths	TOTAL
```

as compared with previous timings

```
JUPYTER_ENTRY_POINT_TIMINGS=1 JUPYTER_ENTRY_POINT_FINDER=entrypoints jupyter --paths
4.02ms jupyter_config_paths loaded with entrypoints
0.27ms	jupyter_config_paths	entry-point-example-setuptools
0.29ms	jupyter_config_paths	entry-point-example-flit
4.63ms jupyter_config_paths	TOTAL
3.49ms jupyter_data_paths loaded with entrypoints
0.09ms	jupyter_data_paths	entry-point-example-setuptools
0.09ms	jupyter_data_paths	entry-point-example-flit
3.70ms jupyter_data_paths	TOTAL
```

File: jupyter_core/paths.py [hunk 3]
```diff
@@ -20,6 +20,7 @@
 import functools
 import traceback
 import importlib
+import importlib_metadata
```
gaborbernat, line 23: You can make this use stdlib on newer pythons as in  https://github.com/pypa/virtualenv/blob/adcf327ee507498a613dc6f235c24c8f2576b6c6/src/virtualenv/run/plugin/base.py#L6-L9

setup.cfg
```diff
@@ -25,6 +25,8 @@ python_requires = >=3.6
 install_requires =
     traitlets
     entrypoints
+    importlib_resources
```
gaborbernat, line 28: You're not using this yet.

setup.cfg
```diff
@@ -24,11 +24,19 @@ include_package_data = True
 python_requires = >=3.6
 install_requires =
     traitlets
+    entrypoints
+    importlib_resources
+    importlib_metadata
```
gaborbernat, line 29: You can also make this optional https://github.com/pypa/virtualenv/blob/adcf327ee5/setup.cfg#L46

setup.cfg
bollwyvl: Sorry, had just woken up, can only do so much before the first pot of coffee.

setup.cfg
bollwyvl: opionalized

gaborbernat:
> > Note the above does not work yet, but I don't see a reason why we could not add support for it.

> (note that `entry_point_example_setuptools.etc.jupyter` is _not_ an importable module, but it is the directory `entry_point_example_setuptools/etc/jupyter`). Then we'd retrieve the absolute path to that directory with importlib_resources.path. Am I understanding you correctly? Can we have an entry point that is not an importable module, and can we ask for a resource that is a directory?

``path`` from ``importlib.resources`` at the moment only supports files, though you can use ``contents`` to unwrap the directory structure. Note in general though python packages are not always installed as exploded (though more often than not they are). For example, zipapps don't contain the files on the disk. In this case, you might want to cache this probably inside an application data folder via ``importlib.resoures.read_bytes`` (this is a one time setup operation for zipapps only). Otherwise, just get the path from the disk. The ``importlib.resources``\``importlib.metadata`` doesn't make this easy to do today, hence pinging the maintainers to see their point of view. In general, though the python packaging/stdlib ecosystem is pushing for ``importlib.resources`` + ``importlib.metadata`` angle. So if we want something supported long term you probably want to go down this path. Some parts in it might not be there yet, but as I said above if something is missing we're open in adding it and jupyter can start using it today via the backport packages.

bollwyvl:
@gaborbernat thanks for the insights! 

> path from importlib.resources at the moment only supports files, though you can use contents to unwrap the directory structure. 

This also triggers an import of the module, which we're trying _very hard_ to avoid. However, I am warming up to the entry_point being a single path, rather than a list, and never being loaded, _and_ the `attr` not even _being_ importable, so we can use the `entry_point` metadata, and not even look at the source. Hopefully I can get my CI up to where it's demonstrating that back to 3.6.

>  python packages are not always installed as exploded

We'd like to handle the zipball case.... someday. It might _possibly_ even improve the windows situation. But in the near term, I'm not sure if we can handle more than the simplest, `zip_safe=False` packages robustly, either "in" `site-packages` (by normal means or editable via `pth` or otherwise) or `PYTHONPATH` (ick... but should test!).

jaraco:
> > As a check on our assumptions and conclusions about python packaging, I've asked @gaborbernat to comment.
> 
> In general here the idea is to use `importlib.metadata` and `importlib.resources`. Both of these are part of the standard library in newer releases, but are also available as backports (the backports mean you can use it today and don't have to wait until the standard library rolls out). See:
> 
>     * https://importlib-resources.readthedocs.io/en/latest
> 
>     * https://importlib-metadata.readthedocs.io/en/latest/using.html#entry-points
> 
> 
> Something like:
> 
> ```python
> from contextlib import ExitStack
> from importlib.metadata import entry_points
> from importlib.resources import path
> 
> stack = ExitStack()
> paths = set()
> for group, eps in entry_points().items():
>     if group == "x.y.z":
>         for ep in eps:
>             path = stack.enter_context(path(ep.module, ep.value))
>             paths.add(path)
> ```
> 

First off, thanks Bernát for sharing these ideas. There's one recent change that's relevant here - the API for `entry_points` is [changing](https://importlib-metadata.readthedocs.io/en/latest/history.html#v3-6-0), so it would be best to use `importlib_metadata>=3.6` (requires Python 3.6 or later) and the newer API (planned to be [included in Python 3.10](https://github.com/python/cpython/pull/24782)):

```
stack = ExitStack()
paths = set(
    stack.enter_context(path(ep.module, ep.value))
    for ep in entry_points(group='x.y.z')
)
```

> Note the above does not work yet, but I don't see a reason why we could not add support for it. ... I'd expect to be able to discover files without needing to import the package.

The reason it needs to import the package is because loading resources relies on the module spec and its loader to implement the resource loading. If the user can't afford the cost of importing the modules, then `importlib.resources` is probably not right for you, and your best bet is to handle the EntryPoint properties in a custom implementation.

> > > Note the above does not work yet, but I don't see a reason why we could not add support for it.
> 
> > (note that `entry_point_example_setuptools.etc.jupyter` is _not_ an importable module, but it is the directory `entry_point_example_setuptools/etc/jupyter`). Then we'd retrieve the absolute path to that directory with importlib_resources.path. Am I understanding you correctly? Can we have an entry point that is not an importable module, and can we ask for a resource that is a directory?
> 
> `path` from `importlib.resources` at the moment only supports files, though you can use `contents` to unwrap the directory structure. Note in general though python packages are not always installed as exploded (though more often than not they are). For example, zipapps don't contain the files on the disk. In this case, you might want to cache this probably inside an application data folder via `importlib.resoures.read_bytes` (this is a one time setup operation for zipapps only). Otherwise, just get the path from the disk. The `importlib.resources```importlib.metadata`doesn't make this easy to do today, hence pinging the maintainers to see their point of view. In general, though the python packaging/stdlib ecosystem is pushing for`importlib.resources`` + ``importlib.metadata`` angle. So if we want something supported long term you probably want to go down this path. Some parts in it might not be there yet, but as I said above if something is missing we're open in adding it and jupyter can start using it today via the backport packages.

The `path` attribute is superseded by the `files()` API (new in `importlib_resources 1.1 ish` and Python 3.9). The `files()` API _does_ support directories as resources through a `Traversable` protocol (very similar to a pathlib object, and actually implemented _as_ a pathlib.Path object for typical filesystem-based packages):

I'd give this a try. First, the entry point definition:

```
[options.entry_points]
jupyter_config_paths =
    entry-point-example-setuptools = entry_point_example_setuptools:etc/jupyter
```

Then load that:

```
eps = entry_points(group='jupyter_config_paths')
ep = eps['entry-point-example-setuptools'] # or iterate over eps
dir = importlib_resources.files(ep.module).joinpath(ep.value)
for file in dir.iterdir():
  text = file.read_text()
  # ... do whatever with text
```

Does that help? This approach would be compatible with zip-eggs, wheels, PYTHONPATH, and other custom loaders that implement the API. If you need files on disk, `importlib.resources` provides `as_file` to create a context around ensuring that any given file is present on the file system (not whole directories, though) for the duration of the context.

bollwyvl:
@jaraco really appreciate you stopping by!

> If the user can't afford the cost of importing the modules, 

Right, our worry is, `jupyter_core` gets used in all kinds of places, some of which might not actually need all the modules imported. Our classic punching bag is `pandas`: want it in your interactive data analysis session? Yes! Want it in the baseline footprint of your web app? Not really....

> ensuring that any given file is present on the file system (not whole directories, though) for the duration of the context.

Because we're catering to other languages, with e.g. `jupyter --paths --json`, we need these directories to exist after the python process `sys.exits`.

I don't know where this puts us, long term: we basically need a `data_files` mechanism, which all existing python-based jupyter-adjacent packages use today, but in a way that works with editable (un)installs, and doesn't duplicate files if we _do_ want to ship them in-tree. And it sounds like unless we:

- require an as-yet-unreleased new dependency 
- implement our own fuse-like filesystem aware of zipballs, e.g. `jupyter_cored`

... the existing `importlib_*` stuff isn't going to move us towards the goal of simplifying packaging static data assets and cross-language config files.

So on this PR: I think shifting _back_ to `entrypoints`, but adopting the single-path convention without parsing, will give python-based jupyter packagers (but ultimately our python and non-python users) the best ergonomics with the least boilerplate... far better, and far less, respectively, than `data_files`, for sure!

jasongrout:
> So on this PR: I think shifting _back_ to `entrypoints`, but adopting the single-path convention without parsing, will give python-based jupyter packagers (but ultimately our python and non-python users) the best ergonomics with the least boilerplate... far better, and far less, respectively, than `data_files`, for sure!

I think you're right, @bollwyvl, though I wonder if we ought to go all the way back to what we had before with entrypoints being python attributes (that we can shortcut to by parsing the module instead of loading it). 

In the end, we need to get from a package a directory (or list of directories) that exist on disk in that package's internal site-packages folder. It has to be a directory, and it has to exist on disk for external programs to use long after the process exits. Furthermore, we want to avoid the cost of importing the package (always if possible, and certainly in the common case). This means that the package is not zip-safe, for one. Summarizing, we have two ideas in play:

A. The package has an entry point that is pointing directly to a directory in its site-packages directory, e.g.,
    ```
    [options.entry_points]
    jupyter_config_paths =
        mypackage = mypackage:etc/jupyter
    ```
    and we write a custom entry point implementation to avoid importing `mypackage`, and to interpret the value as a relative path.
B. The package has an entry point that points to a python attribute loadable , e.g., 
    ```
    [options.entry_points]
    jupyter_config_paths =
        mypackage = mypackage:JUPYTER_CONFIG_PATH
    ```
    where `mypackage/__init__.py` defines `JUPYTER_CONFIG_PATH=['etc/jupyter']`. In this case, we write a custom entry point implementation that attempts to parse the module to avoid the cost of importing, and fall back to importing if we can't parse it.

If we go with option A - are we fighting against what entry points are too much?  It sounds like entry points very much are envisioned as importable python attributes - is that true, or are entry points envisioned as having arbitrary values that can be interpreted however one wants (in our case, a directory on disk)? I don't want to abuse entry points to the point that future developments and tooling around entry points make our usecase very awkward or impossible.

gaborbernat:
I might be mistaken but I think for just iterating over the entry points you don't have to import the distributions. And if at the moment that's not the case I'd expect that to be a bug 😲

bollwyvl:
>  just iterating over the entry points

No, just iterating doesn't trigger the import... it's `path(ep.module, ep.value)`. But the timings suggest _something_ is going on in there for the iterating.

```
26.95ms jupyter_config_paths loaded with importlib_metadata
 4.02ms jupyter_config_paths loaded with entrypoints
```

bollwyvl:
> `mypackage:etc/jupyter`

`flit` doesn't like non-pythonic characters in the target. :cry: 

> are we fighting against what entry points are too much

I think an entrypoint is... whatever the definer of the entrypoint says it is?

I think what's currently pushed is perhaps in the _abuse_ category, if we're counting, but would give:
- `jupyter_*` maintainers a really low number of new moving parts, 
- packagers the tools they need with minimum possible boilerplate... 
  - with the minor caveat of the unfancy path naming/location requirement... 
    - which we want to be as short as possible, anyway. 
- end users 
  - a very stable, shallow dependency
    - which seems appropriate for this layer of the jupyter stack.
  - the minimum possible performance impact we can manage

For the most part, most packages will just need one data path and one config path. If they need more.... they add 1/2 more lines in their declarative packaging file, or `setup.py` if they must... whcih will become static when it goes into a whl. The more i've been thinking about this: if one (well, me) is dynamically messing with the env... I'd just _write to `Path(sys.prefix) / "share/jupyter"`, and not mess with any of this stuff.

gaborbernat:
> >  just iterating over the entry points
> 
> No, just iterating doesn't trigger the import... it's `path(ep.module, ep.value)`. But the timings suggest _something_ is going on in there for the iterating.
> 
> ```
> 26.95ms jupyter_config_paths loaded with importlib_metadata
>  4.02ms jupyter_config_paths loaded with entrypoints
> ```

@jaraco is this a regression 🤔

bollwyvl:
> is this a regression :thinking: 

I may also have been Doing It Wrong, as i was hitting some of the API changes in the docs that i didn't have with 3.9 stdlib... almost certainly,  i was scanning _all_ the `entry_points`, as i couldn't get `group="jcp"` or `select(group="jcp")` to work consistently in different python versions.

jasongrout:
> > `mypackage:etc/jupyter`
> 
> `flit` doesn't like non-pythonic characters in the target. 😢
> 
> > are we fighting against what entry points are too much
> 
> I think an entrypoint is... whatever the definer of the entrypoint says it is?

These two statements seem to conflict with each other. Apparently entry points are also what the package managers envision that they are, i.e., we *are* fighting against things if we are (ab)using them in ways they were not envisioned by the larger community.

jasongrout:
Question for the packaging people on this thread (thanks again *very* much for chiming in here!): Are entry point key/values supposed to be:

A. a nested python attribute inside a loadable module, or
B. arbitrary strings that are interpreted by whoever has defined them and is using them (with extra *optional* support for interpreting them as a nested python attribute inside a loadable module)?

If the vision from the packaging authority is (A), I think we shouldn't abuse them to be (B).

bollwyvl:
> entry points are also what the package managers envision that they are

Yes, as _we_ (ok, _me_, thus far in the code, but _we_ on this PR) are defining what _these_ `entry_points` are, and _they work with `flit`_ is a good litmus for _they works_. 

But again, is _please use the shortest, non-fanciest, most-portable path you can manage_ that bad of a request? I mean, I'd love to ship a `jupyterlab`:rocket:`-extension`, but _somewhere_ that's going to break, anyway (for example, if checked into BitBucket) even if stdlib would allow me to use it and pip would install it.

Anyhow, I should add a CI-exercised `poetry` example, too, but would have to draw the line, somewhere!

gaborbernat:
> Question for the packaging people on this thread (thanks again *very* much for chiming in here!): Are entry point key/values supposed to be:
> 
> A. a nested python attribute inside a loadable module, or
> B. arbitrary strings that are interpreted by whoever has defined them and is using them (with extra *optional* support for interpreting them as a nested python attribute inside a loadable module)?
> 
> If the vision from the packaging authority is (A), I think we shouldn't abuse them to be (B).
> 

I think A by design and use case at the moment, but given the use case I'd personally not be against generalizing to B.

jasongrout:
Reading https://docs.python.org/3/library/importlib.metadata.html#entry-points carefully:

> The group and name are arbitrary values defined by the package author

and https://packaging.python.org/specifications/entry-points/#data-model

> The name identifies this entry point within its group. The precise meaning of this is up to the consumer. For console scripts, the name of the entry point is the command that will be used to launch it. Within a distribution, entry point names should be unique. If different distributions provide the same name, the consumer decides how to handle such conflicts. The name may contain any characters except =, but it cannot start or end with any whitespace character, or start with [. For new entry points, it is recommended to use only letters, numbers, underscores, dots and dashes (regex [\w.-]+).

So another crazy idea: we just use the *name* as the relative directory (it's explicitly an arbitrary string), and basically ignore the value. Apparently we (explicitly) get to decide how to handle conflicts, in which case we just ignore the fact that many things will conflict and add all the paths to our hierarchy.



> Of course, if we slap an `__init__.py` in there, suddenly it _is_ importable 😁 ...

Nick at his finest, finding a way around all objections with a perfectly good idea - brilliant :). Though adopting the "importing things is fine" leads us to importing the top-level package `__init__.py`, which is what we are trying to avoid if possible.

jasongrout:
> I think A by design and use case at the moment, but given the use case I'd personally not be against generalizing to B.

By the way, having arbitrary metadata that could be easily queried and was not attributes of a python module was what I was trying to do with adding extra metadata files at https://github.com/jupyter/jupyter_core/pull/209#issuecomment-791400757 (using https://setuptools.readthedocs.io/en/latest/userguide/extension.html#adding-new-egg-info-files). We eventually gave up on this since it seemed that flit/poetry/etc. doesn't support writing arbitrary files to dist_info.

Brainstorming: perhaps generalizing entry point values to arbitrary strings is what can give us that arbitrary package metadata we want here, or perhaps a new mechanism for arbitrary package metadata that can be queried would be better? Perhaps PKG_INFO is the right place for arbitrary key/value metadata?

jasongrout:
I think it is really important for us to not abuse the entry point system or use it for things it was not intended for (as of now). If we did, we risk future entry point development making our conventions awkward or impossible, and it would just be plain confusing to users, especially users that already understand entry points. I'm assuming here that making entry point values explicitly support arbitrary metadata will be a long process...

Here's what I think we should do: revert back to the version of the code that used entry points the way they were intended to be used, i.e., as importable python attributes, and keep the shortcut of parsing the file for the common case. Test to see how this scales (with, say, 1000 packages). If it's not deemed bad, keep it. We could even have another shortcut to handle Nick's case - if the module to be imported is an empty file (e.g., Nick's `__init__.py` in the actual directory we want), consider the directory of the module as the path we are seeking.

If in the future entry points become a mechanism for arbitrary metadata, we can always add that as well, disambiguating the value as a python attribute or a path by the presence of a `:`.

jupyter_core/paths.py
```diff
@@ -9,21 +9,78 @@
...
 pjoin = os.path.join
 
 # UF_HIDDEN is a stat flag not defined in the stat module.
 # It is used by BSD to indicate hidden files.
 UF_HIDDEN = getattr(stat, 'UF_HIDDEN', 32768)
 
 
+# The entry_point MUST be the name of a directory, consisting of only
+# letters, numbers, and underscores, adjacent to __init__.py
+JUPYTER_DATA_PATH_ENTRY_POINT = "jupyter_data_path"
```
bollwyvl, line 37: on closer reading of the spec, we should probably go with `jupyter_core.data_path` and `jupyter_core.config_path` (or drop the `path` bits), as this is a _new entry point group_.

bollwyvl:
> importable python attributes
> ...
> if the module to be imported is an empty file

I'll see if i can get that together. It doesn't add any non-stdlib dependencies, and if it gives us some wiggle room for the future. One issue with having dotted notation to the left of the `:` is that the top-level module _will_ get [imported](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec):

```
If name is for a submodule (contains a dot), the parent module is automatically imported.
```

So I don't know yet how we might avoid the import behavior... i suppose tossing a `warning` might help folk back onto the happy path of, `my_top_level_name:STATICALLY_PARSEABLE_STRING`, but there's nothing for it with e.g. namespace modules.

> Test to see how this scales 

That'll be more fun :stuck_out_tongue_closed_eyes:

bollwyvl:
Starting lab:
```
jupyter lab --no-browser --debug
# a delay
[D 2021-03-09 21:57:16.340 ServerApp] Searching 
[D 2021-03-09 21:58:16.178 ServerApp] 200 GET /api/contents?content=1&1615345084409 (127.0.0.1) 2.36ms
```
a minute to first pixels isn't too pretty :cry:

bollwyvl:
throwing in a little bit of cache helps immeasurably... well, measurably... but i haven't measured it. 

```python
def _entry_point_paths(ep_group):
    return _cached_entry_point_paths(ep_group, math.floor(time.time() / 100))

@functools.lru_cache(maxsize=10)
def _cached_entry_point_paths(ep_group, epoch):
    ...
```

jasongrout:
> a minute to first pixels isn't too pretty 😢

Ouch. I suppose it does have to open lots of files, which is going to be an even bigger pain on NFS and slower filesystems.

jasongrout:
For completeness in documenting discussions in Jupyter around entry points, see also https://github.com/jupyter/notebook/issues/2894.

jasongrout:
Unfortunately, as far as I can tell, conda does not support general entry points, just `console_script` entry points: https://github.com/conda/conda/issues/9951 (see also https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#python-entry-points). I think this means many of the packages in our ecosystem would have problems using entry points, as they provide official conda-forge builds as well.

bollwyvl:
> conda does not support general entry points

Put back your pitchforks, No worries here! A number of ecosystems (like pytest) would fall entirely apart. The reason `conda-build` handles them explicitly is the default wrappers generated for `console_scripts` are usually broken. So if your package isn't adding a `jupyter <mycmd>` or something, it wouldn't even have that custom entry under `build`.

bollwyvl:
Also: tried the PEP 420 namespace package thing... might be a non-starter as _totally unsurprisingly_ the files wouldn't be in place with a `pth` file (such as on windows) with an editable install.

jasongrout:
> Put back your pitchforks, No worries here!

I was just testing things to see if what conda recipes call "entry points" are in reality just "console_script entry points", and if it just left all other entry points alone that were already in the dist_info directories. 

...and yes, installing a conda environment with `jupyterlab nbconvert vaex` and using importlib_metadata reveals lots of entry points.

Pitchfork being sheathed :).

jasongrout:
FYI @bollwyvl, it looks like to me that if you have many entry points with the same name, the `entrypoints` package gives you back just one, whereas `importlib_metadata` gives you back all of them.

jasongrout:
Here are my timings for `jupyter --paths` after following these instructions from my branch for installing 1000 packages with entry points (so 2000 total entry points): https://github.com/jasongrout/jupyter_core/tree/0310f4a199ba7da60abc54bd9115f7da9a9cec25/examples/scale


Using the entrypoints package:
```
% jupyter --paths > /dev/null
261.42ms jupyter_config_paths loaded
745.81ms jupyter_config_paths	TOTAL
265.96ms jupyter_data_paths loaded
403.42ms jupyter_data_paths	TOTAL
```

Using `importlib_metadata` to get the entry points
```
% JUPYTER_ENTRY_POINT_IMPORTLIB=1 jupyter --paths > /dev/null
427.97ms jupyter_config_paths loaded
944.01ms jupyter_config_paths	TOTAL
417.37ms jupyter_data_paths loaded
561.50ms jupyter_data_paths	TOTAL
```

jaraco:
> Because we're catering to other languages, with e.g. `jupyter --paths --json`, we need these directories to exist after the python process `sys.exits`.

I should perhaps clarify this in the importlib.resources docs. The access to resources on the file system is meant to be for the duration of the context manager and that any expectation of use outside of that should be implemented downstream. In other words, if having a copy after the interpreter exits is a goal, I'd recommend to build a routine that manages that lifecycle and copies the content to the more permanent location. The Python import system has little control over the state of the system between interpreter runs (including pip uninstalls) and there's no proposed spec that I'm aware that would enable management of resources across runs.

> And it sounds like unless we:
> 
>     * require an as-yet-unreleased new dependency
> 
>     * implement our own fuse-like filesystem aware of zipballs, e.g. `jupyter_cored`
> 
> 
> ... the existing `importlib_*` stuff isn't going to move us towards the goal of simplifying packaging static data assets and cross-language config files.

It does seem like `importlib.resources` doesn't implement a solution for this use-case. I should point out that `importlib.resources` supports much more than just the usual FileLoader and ZipLoader, but provides a protocol for other custom loaders (imagine loading modules from a database or from an RPC) to provide resources. Jupyter may not want to support those cases, but if it did, it would need to honor the interface presented by `importlib.resources`.



> I think an entrypoint is... whatever the definer of the entrypoint says it is?

There is a [definition](https://packaging.python.org/specifications/entry-points/#data-model) for entry points and that definition does state that the value should be an importable module and optional name inside that module :/.

It does feel like mild abuse to violate this stated intention.

If there were a clear and obvious way for a package to expose another form of arbitrary metadata, that would be my recommendation, but I'm not sure if such an approach is readily feasible in the current metadata design, as I've not seen it before.

But I just tested it, and I think this could work. Instead of using `entry_points` against their design, provide your own metadata file for your hooks. In [jaraco/develop@demo-metadata-writer](https://github.com/jaraco/jaraco.develop/tree/demo-metadata-writer), I've added a `egg_info.writer` to that project, allowing that project to act as a plugin for setuptools and write additional metadata for any project that includes it during the build.

Then, in the `irc` project (chosen arbitrarily), I include that as a build dependency and demo how the metadata is in fact reachable at run time:

```
irc main $ git diff
diff --git a/pyproject.toml b/pyproject.toml
index b6ebc0b..e6e81dd 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4.1"]
+requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4.1", "jaraco.develop@git+https://github.com/jaraco/jaraco.develop@demo-metadata-writer"]
 build-backend = "setuptools.build_meta"
 
 [tool.black]

irc main $ pip-run -q .
Python 3.9.2 (v3.9.2:1a79785e3e, Feb 19 2021, 09:06:10) 
[Clang 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.

>>> import importlib.metadata as md
>>> md.distribution('irc').read_text('jupyter hooks.txt')
'hook is at foo/bar.py'
``` 

You'd still need a way to solicit the exact hooks for each project. I'd recommend soliciting the hooks from a pyproject.toml, something like

```
[tool.jupyter.hooks]
...
```

In this way, you're following the same principles as setuptools uses to solicit and expose entry points, but you're defining a custom format for a distinct purpose. You would have to design and implement the syntax for the file and parse it yourselves, but you probably want that anyway. The advantage is you have imminent control over the syntax and experience and you're still using the same metadata mechanism as entry points and other packaging patterns.

I'd be willing to help guide this implementation if it sounds attractive.

> @jaraco is this a regression 🤔

Yes, and an intended one with [importlib_metadata 3.5](https://importlib-metadata.readthedocs.io/en/latest/history.html#v3-5-0). Essentially, in order to deduplicate distributions correctly, the metadata for each distribution needs to be loaded. There are plans in python/importlib_metadata#283 to improve performance in light of that concern.

jasongrout:
> I should perhaps clarify this in the importlib.resources docs. The access to resources on the file system is meant to be for the duration of the context manager and that any expectation of use outside of that should be implemented downstream. In other words, if having a copy after the interpreter exits is a goal, I'd recommend to build a routine that manages that lifecycle and copies the content to the more permanent location. The Python import system has little control over the state of the system between interpreter runs (including pip uninstalls) and there's no proposed spec that I'm aware that would enable management of resources across runs.

Thanks for weighing in on this. Interestingly, one of the primary reasons for us to move to entry points over using data_files is that *Python* will manage the lifecycle of these files. Perhaps we're chasing a pipe dream if we need to build something generic enough to support any way a python module might be loaded, but also need the resources to be available outside of Python.

> If there were a clear and obvious way for a package to expose another form of arbitrary metadata, that would be my recommendation, but I'm not sure if such an approach is readily feasible in the current metadata design, as I've not seen it before.

Nice, thanks! This looks like the approach I was attempting in the "Alternative Solutions" section in the issue description, in commit https://github.com/jasongrout/jupyter_core/commit/66351b0978f21b275dd58ab13e6448ef3c610705 (however, I was really fumbling to get the metadata out of the distributions, and I'm sure I made some inaccurate assumptions involving top_level.txt, for example). We decided to abandon this approach in favor of entry_points since various packagers like poetry, flit, etc., don't seem to support arbitrary metadata files, and having broad packager support was one of our design goals.

By the way, I've been thinking over the past few days about how to make finding a specific group of entry points potentially faster (I haven't benchmarked any experiments, so of course this should be treated with appropriate skepticism). It seems that getting a specific group of entry points requires reading in and parsing *all* entry point metadata files in the entire python installation, then filtering for the group I want. My hypothesis is that checking if a file exists is much faster than opening and parsing a file. If each group of entry points was stored in a separate file inside the dist_info/egg directory (for example, as files named by the group in a new `entry_points` directory), it may be much faster to scan for and parse just the data corresponding to a specific entry point group.

gaborbernat:
Worth a try, but my guess is that the largest proportion of the slowness comes from the disk list operation 🤔 but would be great to see some benchmark numbers on discovery vs parsing overhead.

bollwyvl:
Had some other thoughts about our scale issue. And, for reference, a quick look revealed that we are talking about a rough venn diagram of:
- 5,032 candidate packages on pypi that _mention_ jupyter
- 833 candidate npm packages that may already be in pypi, or will be

so these scale concerns are not _entirely_ academic bikeshedding.

Regarding benchmarking: yeah, the above were all with `entrypoints` vs `my_module:foo` and not even checking (much less _importing_) what `foo` was (assuming it to be `my_module/foo`). I didn't even try the exercise with `importlib*`. Things were slightly better, once there was _some_ caching in place, but outside of _just_ the startup issues, a number of other parts of the stack, e.g. `jinja2.FileSystemLoader` and `tornado.web.StaticFileHandler` were never designed to be used with such large search paths.

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/how-could-data-files-be-improved/8972/2

jupyter_core/paths.py
```diff
@@ -9,21 +9,78 @@
...
+# TODO: decide whether to keep, or grow a logging endpoint
+JUPYTER_ENTRY_POINT_TIMINGS = os.environ.get("JUPYTER_ENTRY_POINT_TIMINGS")
+
+
+def _get_path_from_one_entry_point(ep):
+    """ use the entrypoint attribute name to discover the path without loading
+    """
+    spec = importlib.util.find_spec(ep.module_name)
+    module = importlib.util.module_from_spec(spec)
+    origin = pathlib.Path(module.__file__).parent.resolve()
```
telamonian, line 49: `module.__file__` is `None` if there's no top-level `__init__.py` file in the module

telamonian:
I brought this up in the jlab dev call today: https://hackmd.io/Y7fBMQPSQ1C08SDGI-fwtg?both#5-May-2021-Weekly-Meeting

@bollwyvl @jasongrout What's the status of the work on `entry_points`? After reading through this issue and a lot of related stuff, my sense is that no matter what we do, the fact that implementing jupyter `entry_points` requires checking 1000s of paths for plugins/config at runtime is going to cause problems we currently don't have

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/package-managers-extension-paths/11723/2

blink1073:
Closing in favor of using [`shared-data`](https://hatch.pypa.io/latest/plugins/builder/wheel/#options) in hatch.   Thanks @bollwyvl and all for pushing on this front!

PR #213, opened 2021-01-28
[win32_restrict_file_to_user] allow user to delete file

vidartf:
This allows the user to delete the file without elevating / UAC, c.f. https://docs.microsoft.com/en-us/windows/win32/fileio/file-security-and-access-rights

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -396,7 +396,7 @@ def win32_restrict_file_to_user(fname):
 
     dacl = win32security.ACL()
     # dacl.AddAccessAllowedAce(win32security.ACL_REVISION, con.FILE_ALL_ACCESS, everyone)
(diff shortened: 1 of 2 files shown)


vidartf:
I'll have a look at adding a regression test as well.

vidartf:
Xref https://github.com/jupyter/jupyter_client/issues/614

vidartf:
Test added, should be ready for review 👍

kevin-bates (approved):
Great catch Vidar!  Looks like we've been [leaking connection files](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/connect.py#L414-L417) on Windows for a bit.

kevin-bates:
I'd prefer to get Matt's review (requested), then I can cut 4.7.1.

kevin-bates:
I'm going ahead with merging this.  I think Matt is very busy these days and the change makes total sense.

kevin-bates:
@vidartf - 4.7.1 has been released: https://pypi.org/project/jupyter-core/

Issue #211, opened 2021-01-23
RFE: Don't traceback on command not found


hroncok:
Hello. Currently, when a command is not found, Jupter prints a traceback on the user:

```
$ jupyter foobar
Traceback (most recent call last):
  File "/usr/bin/jupyter", line 33, in <module>
    sys.exit(load_entry_point('jupyter-core==4.6.3', 'console_scripts', 'jupyter')())
  File "/usr/lib/python3.9/site-packages/jupyter_core/command.py", line 247, in main
    command = _jupyter_abspath(subcommand)
  File "/usr/lib/python3.9/site-packages/jupyter_core/command.py", line 133, in _jupyter_abspath
    raise Exception(
Exception: Jupyter command `jupyter-foobar` not found.
```

Generally, I believe users should only see tracebacks when they are debugging stuff or when there is an unexpected "crash". An error message formatted for the user is much better UX.

In fact, in Fedora, such traceback is detected as crash and the automatic bug reporting tool:

![_](https://user-images.githubusercontent.com/2401856/105578726-77161680-5d82-11eb-92bd-a36ea6b2ff36.png)

Apparently, many users are hit by this: https://retrace.fedoraproject.org/faf/reports/24959/

Would you accept a PR that changes the exception to printing an error message to stderr instead?

rgbkrk:
Yes, that would be a good PR. Thanks for working on the UX of the command-line tool.

hroncok:
https://github.com/jupyter/jupyter_core/pull/218

PR #218, opened 2021-03-05
Command not found will no longer traceback

hroncok:
Previously, when a command was not found, Jupyter printed a traceback on the user:

    $ jupyter foobar
    Traceback (most recent call last):
      File "/usr/bin/jupyter", line 33, in <module>
        sys.exit(load_entry_point('jupyter-core==4.6.3', 'console_scripts', 'jupyter')())
      File "/usr/lib/python3.9/site-packages/jupyter_core/command.py", line 247, in main
        command = _jupyter_abspath(subcommand)
      File "/usr/lib/python3.9/site-packages/jupyter_core/command.py", line 133, in _jupyter_abspath
        raise Exception(
    Exception: Jupyter command `jupyter-foobar` not found.

Generally, I believe users should only see tracebacks when they are debugging
stuff or when there is an unexpected "crash".

In Fedora, such traceback is detected as crash by the automatic bug reporting tool.

Now, an error message is printed to stderr instead to provide a better UX.

Fixes https://github.com/jupyter/jupyter_core/issues/211

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -282,7 +282,10 @@ def main():
         parser.print_usage(file=sys.stderr)
         sys.exit("subcommand is required")
 
-    command = _jupyter_abspath(subcommand)
+    try:
+        command = _jupyter_abspath(subcommand)
+    except Exception as e:
+        sys.exit(e)
 
     try:
         _execvp(command, sys.argv[1:])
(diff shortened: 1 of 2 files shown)


kevin-bates (approved):
Very nice - thank you.

rgbkrk (approved):
Great, thank you for this fix.

PR #223, opened 2021-04-26
Enable appropriate permissions for CodeQL workflow

kevin-bates:
Adds permissions metadata to `codeql-analysis` workflow since organization-level permissions are read-only.


blink1073 (approved):
Thanks!

Issue #229, opened 2021-05-24
Jupyter eco-system compatibility whish for "pypy3.7-v7.3.5-win64"


stonebig:
Pywin32 doesn't compile on PyPy3 for Windows, as too big/tricky/whatever.

Would it be possible to replace pywin32 for the few functions you call using only  ctypes ?

https://github.com/jupyter/jupyter_core/blob/master/jupyter_core/paths.py#L387.L403

Apparently, it seems the only [known]  blocker to get a jupyter notebook working on "pypy3.7-v7.3.5-win64", [from Jupyter side](https://github.com/jupyter/jupyter_core/issues/226)

stonebig:
I confirm it is the only [remaining] blocker to get a jupyter notebook working on "pypy3.7-v7.3.5-win64"

stonebig:
If anaconda is considering a PyPy3 64 bit on Windows, won't they have the problem too ?

minrk:
closed by #230

PR #230, opened 2021-05-25
Remove dependency on pywin32 package on PyPy

### LINKED ISSUE #226: Remove false dependency on pywin32
There is a stray dependency on pywin32 in `setup.cfg` https://github.com/jupyter/jupyter_core/blob/46a9c512e9e461c848df4a2ea4d2b96c5e527804/setup.cfg#L27

Is it still needed?

cgohlke:
Fix #226 and #229

This PR makes `jupyter_core` compatible with PyPy on Win32.  The main issue is that the `pywin32` package does not yet support PyPy (https://github.com/mhammond/pywin32/issues/1289).

Changes:
1. add an alternative Ctypes implementation of the `win32_restrict_file_to_user` function, which is used as a fallback if the `pywin32` package is not installed
2. change `install_requires` to exclude `pywin32` on PyPy
3. allow `AttributeError: 'stat_result' object has no attribute 'st_file_attributes'` to pass silently

Tested with pypy3.7-v7.3.5.

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -267,8 +267,14 @@ def is_file_hidden_win(abs_path, stat_res=None):
                 return False
             raise
 
-    if stat_res.st_file_attributes & stat.FILE_ATTRIBUTE_HIDDEN:
-        return True
+    try:
+        if stat_res.st_file_attributes & stat.FILE_ATTRIBUTE_HIDDEN:
+            return True
+    except AttributeError:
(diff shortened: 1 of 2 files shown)


stonebig:
is there still a chance to get a jupyter_core-4.7.2 patch release with this in June ?

stonebig:
it would be a small step for jupyter-core, but a great step for PyPy visibility

minrk:
Thank you!

Issue #235, opened 2021-08-18
4.8 Release Plan


jasongrout:
It's been a while since we had a release, and we have a number of good PRs that have been merged. I created a 4.8 milestone with the PRs that have been merged: https://github.com/jupyter/jupyter_core/milestone/20?closed=1

Is there anything else someone feels strongly should go into 4.8? I'm happy to start the release process with a prerelease.

CC @kevin-bates, who did the last several releases.

kevin-bates:
Thank you @jasongrout!

jasongrout:
I tried releasing 4.8.0b0, but it turns out I don't have pypi permissions to release jupyter_core (https://pypi.org/project/jupyter-core/). Can someone give me (`jasongrout`) permissions on pypi to release jupyter_core?

blink1073:
Done!

jasongrout:
> Done!

Thanks. I'm still not showing up in the list of maintainers at https://pypi.org/project/jupyter-core/ (maybe it's a caching issue?)

jasongrout:
> Thanks. I'm still not showing up in the list of maintainers at https://pypi.org/project/jupyter-core/ (maybe it's a caching issue?)

Sorry, I didn't see the email that I needed to accept it first.

jasongrout:
I've released 4.8.0b0. I'll plan to release rc0 early next week if there are no objections. Again, if someone wants to get something else in for 4.8.0, please comment here before next week.

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/jupyter-core-4-8-0-release-candidate/10470/1

jasongrout:
I just released 4.8.0rc0 and posted about the release candidate on Discourse. You can install the RC with `pip install --pre jupyter_core==4.8.0rc0`

Can a few people test it and post here (good or bad)? I propose we release 4.8.0 early next week if we hear only good reports.

blink1073:
I ran `jupyterlab` at [master](https://github.com/jupyterlab/jupyterlab/commit/68a6066dc03dc7cc76141ae78e945faf99da5205) against `jupyter_core==4.8.0rc0` on my 
MacBook Air.  Unit tests pass and the application is fully functional.  I tested the relevant CLI changes as well:

```
jupyter
jupyter help
jupyter --version
jupyter lab --show-config
```

bollwyvl:
[...]

PR #238, opened 2021-09-02
Trigger warning when trying to check hidden file status on PyPy

jasongrout:
As discussed in #235, perhaps the best thing to do to give best-effort support for pypy and help users to understand the current limitations is to trigger a warning when we know we cannot give good results.

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -275,6 +275,7 @@ def is_file_hidden_win(abs_path, stat_res=None):
         # allow AttributeError on PyPy for Windows
         # 'stat_result' object has no attribute 'st_file_attributes'
         # https://foss.heptapod.net/pypy/pypy/-/issues/3469
+        warnings.warn("hidden files are not detectable on this system, so no file will be marked as hidden.")
         pass
 
     return False
--- a/jupyter_core/tests/test_paths.py
+++ b/jupyter_core/tests/test_paths.py
@@ -261,8 +261,8 @@ def test_is_hidden():
 
 
 @pytest.mark.skipif(
-    not (sys.platform == "win32" and "__pypy__" not in sys.modules),
-    reason="only run on windows/cpython: https://foss.heptapod.net/pypy/pypy/-/issues/3469"
+    not (sys.platform == "win32" and (("__pypy__" not in sys.modules) or (sys.implementation.version >= (7, 3, 6)))),
+    reason="only run on windows/cpython or pypy >= 7.3.6: https://foss.heptapod.net/pypy/pypy/-/issues/3469"
 )
 def test_is_hidden_win32_cpython():
     import ctypes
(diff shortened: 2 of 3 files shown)


jasongrout:
@bollwyvl (or others) - can you review this?

stonebig:
isn't this going to solve the problem ? https://foss.heptapod.net/pypy/pypy/-/issues/3469

jasongrout:
> isn't this going to solve the problem ? https://foss.heptapod.net/pypy/pypy/-/issues/3469

It looks like it will in the next pypy release. Having this warning in will still help people that are on older releases, though.

minrk (approved):

jupyter_core/tests/test_paths.py
```diff
@@ -273,6 +274,39 @@ def test_is_hidden_win32():
         assert is_hidden(subdir1, root)
         assert is_file_hidden(subdir1)
 
+@pytest.mark.skipif(
+    not (sys.platform == "win32" and "__pypy__" in sys.modules),
+    reason="only run on windows/pypy: https://foss.heptapod.net/pypy/pypy/-/issues/3469"
```
bollwyvl, line 279: since we're using this twice, should we hoist it to a reusable mark?

bollwyvl: oh, it's _not_ the same w/r/t `not` and `==`! is fine

jupyter_core/paths.py
```diff
@@ -275,6 +275,7 @@ def is_file_hidden_win(abs_path, stat_res=None):
         # allow AttributeError on PyPy for Windows
         # 'stat_result' object has no attribute 'st_file_attributes'
         # https://foss.heptapod.net/pypy/pypy/-/issues/3469
+        warnings.warn("hidden files are not detectable on this system, so no file will be marked as hidden.")
```
bollwyvl, line 278: maybe we have/open a dedicated issue and link to it from here?

bollwyvl (approved):

bollwyvl:
LGTM, modulo the small docs items... but since we anticipate it being fixed upstream soon _anyway_...

jupyter_core/tests/test_paths.py
mattip: maybe add a `and sys.implementation.version < (7, 3, 6)` since after the PyPy 7.3.6 release this should be fixed (nit: there was a bug in `sys.implementation.version` where PyPy used the Python version not the implementation version, but that is all behind us now, and the less-than comparison will still work properly even with the bug)

jasongrout: > maybe add a `and sys.implementation.version < (7, 3, 6)`

Done! Thanks!

jasongrout:
Thanks everyone for your comments and reviews!

Issue #239, opened 2021-09-17
Case-sensitive username on windows test causes failure


jasongrout:
Building the 4.8.0 conda-forge package, we are seeing a test failure on windows:

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=377387&view=logs&j=171a126d-c574-5c8c-1269-ff3b989e923d&t=1183ba29-a0b5-5324-8463-2a49ace9e213&l=1068

```
    def check_user_only_permissions(fname):
        # Windows has it's own permissions ACL patterns
        username = os.environ["USERNAME"]
        permissions = fetch_win32_permissions(fname)
        print(permissions) # for easier debugging
>       assert username in permissions
E       AssertionError: assert 'VssAdministrator' in {'administrators': {'f'}, 'vssadministrator': {'d', 'r', 'w'}}

```

It seems that the fetch_win32_permissions is giving the lowercase "vssadministrator", while the USERNAME environment variable is cased as "VssAdministrator".

Are usernames in windows case-sensitive?

jasongrout:
Well, it looks like we explicitly lowercase the username in the code right above it. 

And it looks like we accidentally removed the lowercasing of the username in this function: https://github.com/jupyter/jupyter_core/commit/2a08a19c430af98934c20da5f3d35396887b755e

Patch coming up to return to lowercasing the username. It does appear from some random googling/duckducking that windows usernames are case insensitive.

jasongrout:
Thanks. Releasing 4.8.1 with this now...

PR #240, opened 2021-09-17
Fix windows test regression.

jasongrout:
Fixes https://github.com/jupyter/jupyter_core/issues/239

It looks like 2a08a19 removed the lowercasing of the username, and in this test the username is compared to the explicitly lowercased username, leading to a test failure on conda-forge.


blink1073 (approved):
Thanks!

PR #242, opened 2021-10-06
Add site user base directory to config and data paths if site.ENABLE_USER_SITE is true

### LINKED ISSUE #221: Should ~/.local/etc/jupyter be part of the default config path?
I apologize if this has been covered before, but I could only find it referenced for a couple specific extensions (https://github.com/mwouts/jupytext/issues/546 and https://github.com/maartenbreddels/ipyvolume/issues/85) and not more generally.  Basically, the problem is that if you `pip install` a Jupyter extension with the `--user` flag, the json file needed to automatically enable the extension ends up somewhere like `~/.local/etc/jupyter/jupyter_notebook_config.d/`, which is not one of the default config directories.

jasongrout:
This allows someone to install Jupyter components in the Python user directory (for example, with `pip install —user`) and Jupyter will see them. It seems like a natural extension of our user-level directories.

Fixes #221 

@minrk - I'm curious what you think of this change.

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -12,6 +12,7 @@ import os
 import sys
 import stat
 import errno
+import site
 import tempfile
 import warnings
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


minrk:
@jasongrout I just had an occasion where I almost made this PR myself, so 👍 from me.

minrk (approved):

vidartf:
@jasongrout I see there was a [comment](https://github.com/jupyter/jupyter_core/issues/221#issuecomment-935657939) about duplicates in data? Is that expected?

vidartf:
AFAICT, the duplicate comes from this logic that is only applied to data dirs:

https://github.com/jupyter/jupyter_core/blob/6de55f54dd84f8d32f534e124c08a6c422f91f34/jupyter_core/paths.py#L90-L102

Does this logic overlap partially/fully with that of `site.getuserbase()` ?

minrk:
Thanks!

PR #248, opened 2021-10-30
set proper sys.argv[0] for subcommand

bnavigator:
Hi,

In https://github.com/openSUSE/libalternatives/issues/11 we noticed that `jupyter subcommand` calls `jupyter-subcommand` with an `argv[0]=subcommand` instead of the proper executable name `jupyter-subcommand`.

Most subcommands do not care, because the first thing they do, is strip `sys.argv[0]` from the parsing:
https://github.com/jupyter/jupyter_core/blob/1778a448ca1c750614371fcf19927cc7b6c6b4db/jupyter_core/application.py#L227-L230

From the Python Doc: [`os.execv*()`](https://docs.python.org/3/library/os.html?highlight=execv#os.execv):
> In either case, the arguments to the child process should start with the name of the command being run, but this is not enforced.

So nothing *mandates* `argv[0]` to be set to the name, but it is good practise and libalternatives relies on it.

You already do the same for subcommands of JupyterApps:
https://github.com/jupyter/jupyter_core/blob/1778a448ca1c750614371fcf19927cc7b6c6b4db/jupyter_core/application.py#L246-L249


jupyter_core/command.py
```diff
@@ -303,7 +303,7 @@ def main():
         sys.exit(e)
 
     try:
-        _execvp(command, sys.argv[1:])
+        _execvp(command, [command] + sys.argv[2:])
```
jasongrout, line 306: Note that `command` here is an absolute path to the command (it was returned by `_jupyter_abspath`). Shouldn't we just pass the base name, i.e., the actual command, rather than the entire absolute path?

jasongrout:
Thanks! Can we also add a test in https://github.com/jupyter/jupyter_core/blob/master/jupyter_core/tests/test_command.py to make sure we don't break this in the future?

jupyter_core/command.py
jasongrout: Never mind - you're right, it *should* be the absolute path.

jasongrout:
@bnavigator - does the test I added look good to you?

bnavigator:
LGTM

PR #249, opened 2021-11-19
Add explicit encoding in open calls

dlukes:
Having enabled `PYTHONWARNDEFAULTENCODING` on Python 3.10, I now get yelled at (which is a good thing!) whenever Python opens a file in text mode without an encoding specified, implicitly falling back on the locale. It just happened when calling `jupyter server --generate-config`, so I took a bit of time to add explicit encoding arguments to open calls in `jupyter_core`.

Please review if UTF-8 is the appropriate encoding in all cases, I'm not otherwise familiar with the codebase, so it's not always easy for me to see. Most of the calls seem to handle Python / CSS / JS files, where UTF-8 should be appropriate, but one of the functions I'm not entirely sure about is `paths.secure_write`, which can probably be used to write lots of different files (though that in turn means it can be used to write Python files, which should be UTF-8).

### DIFF
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -147,7 +147,7 @@ class JupyterApp(Application):
             config_text = config_text.decode('utf8')
         print("Writing default config to: %s" % config_file)
         ensure_dir_exists(os.path.abspath(os.path.dirname(config_file)), 0o700)
-        with open(config_file, mode='w') as f:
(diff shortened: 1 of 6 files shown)


vidartf:
This looks like a clear improvement to me.

jasongrout:
Looks good to me too. Thanks!

PR #251, opened 2022-02-15
jupyter_config_dir - reorder home_dir initialization

dharmaquark:
This PR moves the initialization of the home_dir variable, so that this logic will not get hit if `JUPYTER_CONFIG_DIR` or `JUPYTER_NO_CONFIG` are set in the environment (this helps to avoid any OS level issues accessing or resolving default file paths when the config directory path is already being set explicitly as an environment variable)

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -62,14 +62,13 @@ def jupyter_config_dir():
     """
 
     env = os.environ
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thanks @dharmaquark!

PR #254, opened 2022-04-06
Add project URLs to setup.cfg

tlinhart:
Adding project URLs (especially source code URL) makes it easier for dependency management tools (like Renovate).


blink1073 (changes requested):
Thanks!  Miind adding `"Funding": "https://numfocus.org/",` as well?

blink1073 (approved):
Thanks!

PR #255, opened 2022-04-07
Set up pre-commit

blink1073:


.github/workflows/codeql-analysis.yml
```diff
@@ -14,16 +14,15 @@ name: "CodeQL"
...
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [ master ]
+    branches: [master]
   schedule:
-    - cron: '41 6 * * 6'
+    - cron: "41 6 * * 6"
```
kevin-bates, line 22: I don't have the crontab format memorized and I believe GH scheduled actions are UTC-based, so would it be helpful to add a comment like:
```suggestion
    # Make a pass every Saturday at 06:41 UTC
    - cron: "41 6 * * 6"
```

kevin-bates (approved):
This looks good.  Since I always have to refresh my memory on crontab formats, I just left the one (non-essential) comment.
Thanks for making the build infrastructures consistent throughout!

blink1073:
GitHub says it is fully operational but page loads and actions are definitely degraded

PR #271, opened 2022-06-22
is_hidden: Use normalized paths

martinRenou:
The current `is_hidden` implementation considers the path `path/to/my/./file.txt` to be hidden, while it's not. This PR fixes it by normalizing the paths prior to compute the `is_hidden` logic.

Similar fix in `jupyter_server` https://github.com/jupyter-server/jupyter_server/pull/886


blink1073 (approved):
Thanks!

PR #286, opened 2022-08-22
Try to detect if we are in a virtual environment and change path precedence accordingly

### LINKED ISSUE #283: Make  JUPYTER_PREFER_ENV_PATH=1 default
I propose that `JUPYTER_PREFER_ENV_PATH=1` be made the default behavior.

I cannot imagine a case where someone would set up a virtual environment and not expect the python based software such as Jupyter lab to use what is installed in the virtual environment first. Thus, I believe Jupyter lab should never use versions of extensions, etc outside of the virtual environment in preference to those inside. I can imagine people wanting to set up some utilities that work in all environments. So, can see there might be issues surrounding that. I suggest the following logic:

1. If a package is installed in the virtual environment use that in preference to any other available versions.
2. If a package does not exist in the virtual environment but is available at the user level use it in preference to versions available at the system level.
3. Only use versions available at the system level if they have not been overridden at the previous two levels.
4. There may need to be a switch to not use packages from the user or system level.

I encountered an unexpected issue with ipywidgets because of this (see https://github.com/jupyter-widgets/ipywidgets/issues/3559).

The original implementation was discussed in https://github.com/jupyter/jupyter_core/pull/199

Thanks for a great tool.

Jonathan

jasongrout:
We make the assumption that if we are in a virtual environment, we should prioritize the environment-level sys.prefix over the user-level paths. The user can opt out of this behavior by setting JUPYTER_PREFER_ENV_PATH, which takes precedence over our autodetection.

Fixes #283

### DIFF
--- a/jupyter_core/command.py
+++ b/jupyter_core/command.py
@@ -231,13 +231,13 @@ def main():
                 if args.debug:
                     env = os.environ
 
-                    if paths.envset("JUPYTER_PREFER_ENV_PATH"):
+                    if paths.prefer_environment_over_user():
                         print(
-                            "JUPYTER_PREFER_ENV_PATH is set, making the environment-level path preferred over the user-level path for data and config"
+                            "JUPYTER_PREFER_ENV_PATH is set to a true value or we detected a virtual environment, making the environment-level path preferred over the user-level path for data and config"
                         )
                     else:
                         print(
-                            "JUPYTER_PREFER_ENV_PATH is not set, making the user-level path preferred over the environment-level path for data and config"
+                            "JUPYTER_PREFER_ENV_PATH is not set and we did not detect a virtual environment, or JUPYTER_PREFER_ENV_PATH is set to a false value, making the user-level path preferred over the environment-level path for data and config"
                         )
 
                     # config path list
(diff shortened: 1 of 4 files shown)


jasongrout:
Of course, several tests fail. @gutow (or anyone else), feel free to take over this if you can.

gutow:
I am trying to pick this up. But am having trouble setting up an environment where I can run tests locally.

It looks to me like line 29 is causing the error in the CI testing:
```
def envset(name, default=False):
```
Should be: 
```
def envset(name, default="no"):
```
To be consistent with the vocabulary used previously.

I get errors on running the tests that appear to be caused by isolating the tests in a virtual environment. They are not the same as occur in the github CI testing.

So, I think I would have to set up a virtual machine with no virtual environments to do the tests on my local machine. I have lots of different projects on my machine, so have them all isolated in virtual environments. I definitely do not want to set this one up outside of a virtual environment. Am I on the right track here?

jasongrout:
> It looks to me like line 29 of paths.py is causing the error in the CI testing:

Actually, most of the errors came from an actual bug in my code. Doh! https://github.com/jupyter/jupyter_core/pull/286/commits/8b5f1c5598950ad4890f0daca3deebdeeac9ea08

blink1073 (approved):
LGTM, thanks!

gutow:
Thanks you for this change. I look forward to the next release. I think this will help avoid problems for naive user such as myself.

PR #291, opened 2022-09-26
Add current working directory as first config path

### LINKED ISSUE #228: Jupyter --paths incorrect output
Hi,

`jupyter --paths` outputs:
```
$ jupyter --paths
config:
    /path/to/user/.jupyter
    /path/to/venv/etc/jupyter
    /usr/local/etc/jupyter
    /etc/jupyter
data:
    /path/to/user/.local/share/jupyter
    /path/to/venv/share/jupyter
    /usr/local/share/jupyter
    /usr/share/jupyter
runtime:
    /path/to/user/.local/share/jupyter/runtime
```

Surprisingly, if you have a file named `/path/to/user/jupyter_notebook_config.py` it get's used too! I'm not sure if this is a bug in the output of `--paths` or a bug in which files should be used for config (or if this a bug in jupyter_server).

blink1073:
Fixes #228

### DIFF
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -88,8 +88,8 @@ class JupyterApp(Application):
     def config_file_paths(self):
         path = jupyter_config_path()
         if self.config_dir not in path:
-            path.insert(0, self.config_dir)
-        path.insert(0, os.getcwd())
+            # Insert config dir after cwd.
+            path.insert(1, self.config_dir)
         return path
 
     data_dir = Unicode()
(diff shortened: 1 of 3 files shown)


jasongrout:
Thanks!

kevin-bates:
I just installed the pre-release and am finding that `lab/workspaces` is being _created_ in my home directory, abandoning the same already in `~/.jupyter`.  With the switch to [`platformdirs`](https://github.com/jupyter/jupyter_core/pull/292), there's an opt-in element for the 5.x release.  Is this worthy of the same?  Otherwise, folks will "lose" their current settings and those settings are now visible (by default) if their `root-dir` happens to be their home directory.

PR #292, opened 2022-09-26
Use platformdirs for path locations

### LINKED ISSUE #234: Please consider using appdirs
[appdirs](https://pypi.org/project/appdirs/) takes care of OS dependent directories.  Using it would be a small change to [this file](https://github.com/jupyter/jupyter_core/blob/2dc34f70806b6668f965bfaa39623143cb2cda59/jupyter_core/paths.py#L56).

blink1073:
Fixes #234

### DIFF
--- a/jupyter_core/tests/test_paths.py
+++ b/jupyter_core/tests/test_paths.py
@@ -35,7 +35,7 @@ pjoin = os.path.join
 
 
 xdg_env = {
-    "USER": os.environ["USER"],
+    "HOME": os.environ["HOME"],
     "XDG_CONFIG_HOME": "/tmp/xdg/config",
     "XDG_DATA_HOME": "/tmp/xdg/data",
     "XDG_RUNTIME_DIR": "/tmp/xdg/runtime",
@@ -44,7 +44,7 @@ xdg = patch.dict("os.environ", xdg_env)
 no_xdg = patch.dict(
     "os.environ",
     {
-        "USER": os.environ["USER"],
+        "HOME": os.environ["HOME"],
         "XDG_CONFIG_HOME": "",
         "XDG_DATA_HOME": "",
         "XDG_RUNTIME_DIR": "",
@@ -54,7 +54,7 @@ no_xdg = patch.dict(
 no_config_env = patch.dict(
     "os.environ",
     {
-        "USER": os.environ["USER"],
+        "HOME": os.environ["HOME"],
         "JUPYTER_CONFIG_DIR": "",
         "JUPYTER_DATA_DIR": "",
         "JUPYTER_RUNTIME_DIR": "",
--- a/jupyter_core/tests/test_paths.py
+++ b/jupyter_core/tests/test_paths.py
@@ -92,11 +92,11 @@ def test_envset():
 
 @pytest.mark.skipif(sys.platform != "darwin", reason="only run on osx")
 def test_config_dir_darwin():
-    with darwin, no_config_env:
+    with no_config_env:
         config = jupyter_config_dir()
     assert config.startswith(os.path.expanduser("~"))
 
-    with darwin, config_env:
+    with config_env:
         config = jupyter_config_dir()
     assert config == jupyter_config_env
 
@@ -139,7 +139,7 @@ def test_data_dir_darwin():
 @pytest.mark.skipif(sys.platform != "win32", reason="only run on windows")
 def test_data_dir_windows():
     data = jupyter_data_dir()
-    assert data == realpath(pjoin(os.environ.get("APPDATA", ""), "jupyter"))
+    assert data.endswith(r"AppData\Local\Jupyter\data")
 
 
 @pytest.mark.skipif(sys.platform != "linux", reason="only run on linux")
@@ -169,7 +169,7 @@ def test_runtime_dir_darwin():
 @pytest.mark.skipif(sys.platform != "win32", reason="only run on windows")
 def test_runtime_dir_windows():
     runtime = jupyter_runtime_dir()
-    assert runtime == realpath(pjoin(os.environ.get("APPDATA", ""), "jupyter", "runtime"))
+    assert runtime.endswith(r"AppData\Local\Jupyter\data\runtime")
 
 
 @pytest.mark.skipif(sys.platform != "linux", reason="only run on linux")
--- a/jupyter_core/tests/test_paths.py
+++ b/jupyter_core/tests/test_paths.py
@@ -233,7 +233,7 @@ def test_jupyter_path():
     system_path = ["system", "path"]
     with no_config_env, patch.object(paths, "SYSTEM_JUPYTER_PATH", system_path):
         path = jupyter_path()
-    assert path[0] == jupyter_data_dir()
+        assert path[0] == jupyter_data_dir()
     assert path[-2:] == system_path
 
 
@@ -258,7 +258,7 @@ def test_jupyter_path_user_site():
 def test_jupyter_path_no_user_site():
     with no_config_env, patch.object(site, "ENABLE_USER_SITE", False):
         path = jupyter_path()
-    assert path[0] == jupyter_data_dir()
+        assert path[0] == jupyter_data_dir()
     assert path[1] == paths.ENV_JUPYTER_PATH[0]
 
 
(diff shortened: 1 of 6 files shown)


blink1073:
I don't think we can make this change, I'm running into https://github.com/ActiveState/appdirs/issues/167.


```
  File "C:\hostedtoolcache\windows\Python\3.10.7\x64\lib\site-packages\appdirs.py", line 134, in site_data_dir

    path = os.path.normpath(_get_win_folder("CSIDL_COMMON_APPDATA"))

  File "C:\hostedtoolcache\windows\Python\3.10.7\x64\lib\site-packages\appdirs.py", line 481, in _get_win_folder_with_pywin32

    dir = shell.SHGetFolderPath(0, getattr(shellcon, csidl_name), 0, 0)

pywintypes.com_error: (-2147024893, 'The system cannot find the path specified.', None, None)
```

blink1073:
Looks like `platformdirs` is a drop-in replacement and is maintained.

jasongrout:
> Looks like `platformdirs` is a drop-in replacement and is maintained.

And maintained by Bernat, at that. 👍

jasongrout:
How should we manage the transition? I can imagine this will be a big pain to people that have configs in our hardcoded places. At the very least, it seems like we should have a deprecation period for the old directories. Probably even better is to have an opt-in to the new behavior on this next major version (with deprecation of old directories), and an opt-out in the major version after that, then final removal of old directories in the major version after that?

blink1073:
> Probably even better is to have an opt-in to the new behavior on this next major version (with deprecation of old directories), and an opt-out in the major version after that, then final removal of old directories in the major version after that?

Yeah, that sounds reasonable.

jasongrout:
To elaborate more, here's a proposal: we have a `JUPYTER_PLATFORM_DIRS` environment variable (or a better name?)? If JUPYTER_PLATFORM_DIRS is truthy, we apply this new logic, and if it is falsey we apply the old logic.

- In version 5, if JUPYTER_PLATFORM_DIRS is not defined, we assume it is falsey. We also introduce a deprecation warning when using the old logic. (However, there is a question about how we might make sure users get the deprecation warning. We might use logic similar to what we introduced in ipywidgets to bubble the deprecation from https://github.com/jupyter-widgets/ipywidgets/pull/3569)
- In version 6, we flip the default so that if JUPYTER_PLATFORM_DIRS is not defined, we assume it is truthy
- In version 7, we remove the environment variable checks and old logic.

jasongrout:
@blink1073 - I pushed a commit consolidating the environment checking to a single function that also uses our envset utility function.

jupyter_core/tests/test_paths.py
```diff
@@ -139,39 +139,50 @@ def test_data_dir_env():
...
     assert data == pjoin(xdg_env["XDG_DATA_HOME"], "jupyter")
 
+    with patch.dict("os.environ", {"JUPYTER_PLATFORM_DIRS": "1"}):
+        data = jupyter_data_dir()
+    assert data == realpath("~/.local/share/jupyter")
+
+    with patch.dict("os.environ", {"JUPYTER_PLATFORM_DIRS": "1", "XDG_DATA_HOME": "/tmp/xdg/data"}):
+        data = jupyter_data_dir()
+    # TODO: Should this be titlecase "Juptyer" by xdg convention?
+    assert data == pjoin(xdg_env["XDG_DATA_HOME"], "jupyter")
```
blink1073, line 184: Based on this list, I think it could go either way: https://wiki.archlinux.org/title/XDG_Base_Directory

jasongrout: I think we ought to make it lowercase, then, for backwards compatibility, and to simplify that on linux, it is lowercase.

blink1073: Makes sense to me

jasongrout:
When you want to pause for a bit, there are a few more cleanups of the test file I'd like to do (convenience decorators and context managers for the platforms and platformdirs patch, etc.).

blink1073:
> When you want to pause for a bit, there are a few more cleanups of the test file I'd like to do (convenience decorators and context managers for the platforms and platformdirs patch, etc.).

I'm done with my changes

jupyter_core/tests/test_paths.py
```diff
@@ -99,159 +99,211 @@ def test_envset():
...
+def test_data_dir_windows_legacy():
     data = jupyter_data_dir()
     assert data == realpath(pjoin(os.environ.get("APPDATA", ""), "jupyter"))
 
+
+@windows
+@use_platformdirs
+def test_data_dir_windows():
+    data = jupyter_data_dir()
+    assert data == realpath(pjoin(os.environ.get("LOCALAPPDATA", ""), "Jupyter"))
```
jasongrout, line 179: Is there a clear reason why we should switch windows from APPDATA (which in our tests is AppData/Roaming) to the local appdata (AppData/Local)? We can switch back to remote with the platformdirs remote=True argument.

I don't know Windows well enough to know if there is a strong clear reason we should move to local appdata over just generic appdata.

jasongrout: https://learn.microsoft.com/en-us/windows/apps/design/app-settings/store-and-retrieve-app-data gives some background, including the fact that roaming data is no longer supported as of Windows 11, apparently. So there's a strong reason to move to local appdata. Also roaming data has a pretty severe quota apparently.

jupyter_core/utils/__init__.py
```diff
@@ -15,3 +22,62 @@ def ensure_dir_exists(path, mode=0o777):
...
+    # Return the stack level from the perspective of whoever called us (i.e., one level up)
+    return level - 1
+
+
+def deprecation(message, internal="jupyter_core/"):
+    """Generate a deprecation warning targeting the first frame that is not 'internal'
+
+    internal is a string or list of strings, which if they appear in filenames in the
+    frames, the frames will be considered internal. Changing this can be useful if, for examnple,
+    we know that ipywidgets is calling out to traitlets internally.
```
jasongrout, line 74:
```suggestion
    we know that our internal code is calling out to another library.
```

jasongrout (approved):

jasongrout:
@blink1073 - this all looks good to me now. If it looks good to you, I think we can merge. Thanks!

blink1073:
Great!  I'll cut a 5.0 rc tomorrow

dlqqq:
This change is causing CI failures in my branches. Is the deprecation warning intended to raise an exception?

Build log: https://github.com/jupyter-server/jupyter_server_fileid/actions/runs/3222840389/jobs/5272324591

dlqqq:
I was able to resolve the failures by explicitly setting `JUPYTER_PLATFORM_DIRS=1` in my CI environment. https://github.com/jupyter-server/jupyter_server_fileid/pull/8/commits/941ab674e68eb6747d4cf15fa7747699a4695e3d

kevin-bates:
Hi @dlqqq - your [`pyproject.toml`](https://github.com/jupyter-server/jupyter_server_fileid/blob/54cf04cb84522d643eec483870796e2f261d7f46/pyproject.toml#L52) treats warnings encountered in pytest as errors.

Using [jupyter-server as an example](https://github.com/jupyter-server/jupyter_server/blob/ece308bd9a18fba752b60ab15dead861413c5254/pyproject.toml#L149), this particular DeprecationWarning can be tolerated by adding it to the filters. 
(At least that's my understanding of things.)

dlqqq:
@kevin-bates Ah, you're right. So, `warnings.warn()` only fails in Pytest and not at runtime? Seems very odd, and I can't find this behavior documented anywhere.

kevin-bates:
Yeah, I had to dig a bit, but found [this](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#controlling-warnings) and remember @blink1073 mentioning something about taking a stronger stance regarding warnings in one of his updates. 

If you remove the `error` entry from the [`filterwarnings`](https://docs.python.org/3/library/warnings.html#the-warnings-filter) entry, `pytest` will succeed as well (and display the deprecation warning following the tests).

blink1073:
Yep, that's right, see for example how [jupyter_server](https://github.com/jupyter-server/jupyter_server/blob/1957aa0416c8467961198fa2dd6f930aca106f85/pyproject.toml#L140) handles warnings in tests.

PR #300, opened 2022-10-27
Clean up typings and bump supported python versions

blink1073:


blink1073:
Will redo once https://github.com/ipython/traitlets/pull/788 lands

PR #306, opened 2022-11-08
Add dependabot

blink1073:


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  # Set update schedule for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every weekday
+      interval: "weekly"
```
kevin-bates, line 8: nit: This comment doesn't match the value. Should the comment be once a week or the value be "daily"?

blink1073: Haha cargo culting at it's finest.  I have committed this to about a dozen repos already.

kevin-bates: Sorry, I had responded from my phone earlier and didn't realize this was already merged.  Sounds like "weekly" is what we're going with.  

Thanks for driving the automation for practically the entire organization!

PR #323, opened 2022-12-13
Only prefer envs owned by the current user

### LINKED ISSUE #318: Feature request: do not prefer env path if it is not writeable
Jupyter_core 5.0.0 made the env path preferred (unless you set `JUPYTER_PREFER_ENV_PATH=no`).  This is probably the right general case, but it didn't work for us, and here's why.

We are indeed running that lab from inside an environment (specifically rubin-env-rsp from conda-forge)...but the way we have the Rubin Science Platform constructed, that path is not writeable.  It's read-only within the container.

I suspect you didn't consider that case; it is after all pretty unusual.  I feel like a better default would be to see whether the env path is writeable, and if it is NOT, do not prefer it, but if it is, then prefer it.  This would give the advantages of the 5.0 strategy, but also accommodate our use-case and that of sites doing the same thing we are of providing a user-immutable reference platform.

I don't know if I will have the cycles myself to produce a PR that will implement this, but if I do, I will.

minrk:
Check ownership of prefix to avoid misidentifying 'shared' environments that can't be.

Heuristic being to prefer _user_ environments to general user config, but not _shared_ environments that may be owned by e.g. root.

in cases of a shared virtualenv, the env may be owned by root and not writable, causing problems for e.g. juptyerlab which writes workspaces to the default config dir, which will fail with a permission error. It makes sense to prefer the env when they are my envs for projects, but not for e.g. creating a user env to be used in jupyterhub. I think an ownership  <del>writability</del> check is a good indicator of whether it's a 'shared' env.

More info: https://github.com/jupyterlab/jupyterlab/issues/13580 

This would fix the above issue in the real cases I've encountered, but it's not a rigorous check for the writability of the config dir specifically.

Downside of just a writability check: if an admin has write access instead of requiring e.g. root, they will still get the shared env!

With an owner check, this can still happen if a 'shared' environment is owned by one of its users. That ought to be more rare than _writable_, at least.

closes #318


minrk:
This now has an ownership check, which falls back on W_OK if no owner can be established, which I _think_ will happen on Windows.

blink1073:
@jasongrout had brought up some concerns about special-casing the "writability" of the path [here](https://github.com/jupyter/jupyter_core/issues/318#issuecomment-1330543591).

minrk:
@jasongrout thanks! I left comments in #318. I think shared env detection is too complex and fraught with wrong conclusions (e.g. admin vs user getting different results for the same env). I think the solution probably belongs in checks in jupyterlab  whatever package is choosing to require writing. I do believe it shouldn't be on the user to solve, though, for standard persistence behavior like jupyterlab workspaces.

minrk:
Reopening now that it's switched to an ownership check, following discussion in #318

blink1073 (approved):
Thanks!

PR #324, opened 2022-12-14
Don't treat the conda root env as an env

minrk:
For the purposes of 'preferring env config', I don't think it makes sense to treat the conda root prefix as an environment. This is quite likely to be a shared, not writable installation, and it would be more consistent to be treated like any other base Python install.

related: #318


jasongrout:
If I download the anaconda or miniconda distribution, install it in my user directory, and fire it up, the default is to be in the CONDA_ROOT environment, right? Isn't that one of the most common cases for using anaconda?

jupyter_core/paths.py
```diff
@@ -79,7 +79,11 @@ def prefer_environment_over_user() -> bool:
         return True
 
     # If sys.prefix indicates Python comes from a conda/mamba environment, default to True
```
jasongrout, line 81:
```suggestion
    # If sys.prefix indicates Python comes from a conda/mamba environment that is not the root environment, default to True
```

minrk:
I think CONDA_ROOT may be years old and no longer used (it's in my bashrc). I cleared all conda envs and started fresh:

```
unset $(env | grep CONDA | cut -d= -f1)
source $PREFIX/etc/profile.d/conda.sh
conda activate base
env | grep CONDA
```

```
CONDA_EXE=/Users/minrk/conda/bin/conda
CONDA_PREFIX=/Users/minrk/conda
CONDA_PROMPT_MODIFIER=(base)
_CE_CONDA=
CONDA_SHLVL=1
CONDA_PYTHON_EXE=/Users/minrk/conda/bin/python
CONDA_DEFAULT_ENV=base
```

So the two logical choices are:

1. `CONDA_DEFAULT_ENV=base` (base is a reserved name, you can't create another env with this name)
2. resolve prefix from CONDA_EXE, which is platform-dependent

I elected to consider CONDA_DEFAULT_ENV being _unset_ to mean we're not in an env, because that should mean that `conda activate` has not been called. I could go either way on that, but I went with lacking confidence that we _are_ in an env should mean we don't opt-in to env behavior, rather than the other way around.

blink1073:
I'm happy to try out this and #323 and wait for feedback.

PR #326, opened 2022-12-22
Maint: Some more precise typing.

Carreau:
Mostly make sure that ensure_async is properly typed, and not justs has "Any"

### DIFF
--- a/jupyter_core/utils/__init__.py
+++ b/jupyter_core/utils/__init__.py
@@ -10,8 +10,9 @@ import sys
 import threading
 import warnings
 from pathlib import Path
-from typing import Awaitable, Callable, Optional, TypeVar, Union, List, cast
 from types import FrameType
+from typing import Awaitable, Callable, List, Optional, TypeVar, Union, cast
+
 
 def ensure_dir_exists(path, mode=0o777):
     """ensure that a directory exists
--- a/jupyter_core/utils/__init__.py
+++ b/jupyter_core/utils/__init__.py
@@ -10,8 +10,8 @@ import sys
 import threading
 import warnings
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


jupyter_core/utils/__init__.py
```diff
@@ -163,18 +167,25 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
-async def ensure_async(obj: Union[Awaitable[Any], Any]) -> Any:
+# Matthias I believe this is technically incorrect,
```
blink1073, line 170: Was this meant to stay here as commentary?

Carreau: Removed. I'm still unsure about the correctness.

blink1073: I think we need it to be any awaitable since we use it for Futures too.

Carreau: Yeah, I don't think that's a big deal, but if it were used in a context where a async function return an awaitable on purpose then it's going then it may do the wrong things, 

I thing that 

`ensure_async(fun(val))` should likely be `ensure_async_func(func)(val)`, so you test the function, and not it's return value. I don't think it matters in how we are using it, but I think it's not the best abstraction.

blink1073: We could use `overload` to handle the different  cases

blink1073: But yeah, we should have changed the API to something better when we moved it to core

blink1073: Actually, no, this function is not meant to accept functions, only the result of calling a function, which could either be an awaitable or not.

blink1073 (approved):
Thank you!

PR #330, opened 2023-01-05
MAINT: Coro is not coro-function.

Carreau:
in the following

```python
async def hello():
    return 1
```

`hello` is a coroutine function, and `hello()` a coroutine. Fix the vocabulary and ensure that the function is properly used.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,3 +39,4 @@ repos:
     hooks:
       - id: ruff
         args: ["--fix"]
+        exclude: script
--- a/jupyter_core/utils/__init__.py
+++ b/jupyter_core/utils/__init__.py
@@ -137,14 +137,16 @@ def run_sync(coro: Callable[..., Awaitable[T]]) -> Callable[..., T]:
 
     Parameters
     ----------
-    coro : coroutine
-        The coroutine to be executed.
+    coro : coroutine-function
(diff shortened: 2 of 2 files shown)


davidbrochart (commented):
I think the vocabulary should be `coroutine` and `async function`. `coroutine-function` is confusing.

Carreau:
`coroutine function` is the official term. 

See inspect module [iscoroutinefunction](https://docs.python.org/3/library/inspect.html#inspect.iscoroutinefunction) and [iscoroutine](https://docs.python.org/3/library/inspect.html#inspect.iscoroutine).

Async-def function is ambiguous as it could also be an [async-generator-function](https://docs.python.org/3/library/inspect.html#inspect.isasyncgenfunction), which when called would be an [async-generator](https://docs.python.org/3/library/inspect.html#inspect.isasyncgen)

I mean those things have known, defined meaning, I don't see why we should use a different terminology here.

If you really insist we can change, but coroutine is really the instantiated object, so I'd like not to use it.

davidbrochart:
Thanks, "coroutine function" seems appropriate then.

Carreau:
I'm going to close and reopen to trigger CI.

blink1073 (approved):
Thanks!

blink1073:
The check release failure is unrelated, I'll fix it separately.

PR #333, opened 2023-01-07
Clear the test environment before each function run

jasongrout:
See #231

This attempts to make our test environment more reproducible by setting the environment variables to a known state before each function run.


blink1073 (approved):
Thanks!

PR #353, opened 2023-05-25
go back to using nest_asyncio for run_sync utility

smacke:
This PR appears to fix the issue described in https://github.com/jupyter/notebook/issues/6721, whereby users sporadically encounter `zmq message arrived on closed channel` errors in the tornado loop. It reverts the threaded implementation of `jupyter_core.utils.run_sync` and replaces it with the original one based on `nest_asyncio`.

Actually it uses a patched version of `nest_asyncio` from [here](https://github.com/ipyflow/ipyflow/blob/8e4bc5cb8d4231b9b69f4f9dce867b8101164ac5/core/ipyflow/kernel/patched_nest_asyncio.py), due to portability issues with the original version.

Unfortunately I don't fully understand why reverting the threaded implementation of `run_sync` is needed to eliminate these errors, but some users suggest it could have something to do with tornado coroutines not playing nicely when run across different threads: https://github.com/tornadoweb/tornado/issues/2871#issuecomment-638577266

Test plan: did an editable install locally and verified everything works well, and that all the zmq errors I encountered before are gone.

### DIFF
--- a/jupyter_core/utils/__init__.py
+++ b/jupyter_core/utils/__init__.py
@@ -2,12 +2,10 @@
 # Distributed under the terms of the Modified BSD License.
 
 import asyncio
-import atexit
 import errno
 import inspect
 import os
 import sys
-import threading
 import warnings
 from pathlib import Path
 from types import FrameType
@@ -92,43 +90,6 @@ def deprecation(message: str, internal: Union[str, List[str]] = "jupyter_core/")
 T = TypeVar("T")
 
 
-class _TaskRunner:
-    """A task runner that runs an asyncio event loop on a background thread."""
-
-    def __init__(self):
-        self.__io_loop: Optional[asyncio.AbstractEventLoop] = None
-        self.__runner_thread: Optional[threading.Thread] = None
-        self.__lock = threading.Lock()
-        atexit.register(self._close)
-
-    def _close(self):
-        if self.__io_loop:
-            self.__io_loop.stop()
-
-    def _runner(self):
-        loop = self.__io_loop
-        assert loop is not None  # noqa
-        try:
-            loop.run_forever()
-        finally:
-            loop.close()
-
-    def run(self, coro):
-        """Synchronously run a coroutine on a background thread."""
-        with self.__lock:
-            name = f"{threading.current_thread().name} - runner"
-            if self.__io_loop is None:
-                self.__io_loop = asyncio.new_event_loop()
-                self.__runner_thread = threading.Thread(target=self._runner, daemon=True, name=name)
-                self.__runner_thread.start()
-        fut = asyncio.run_coroutine_threadsafe(coro, self.__io_loop)
-        return fut.result(None)
-
-
-_runner_map = {}
-_loop_map = {}
-
-
 def run_sync(coro: Callable[..., Awaitable[T]]) -> Callable[..., T]:
     """Wraps coroutine in a function that blocks until it has executed.
 
@@ -147,23 +108,24 @@ def run_sync(coro: Callable[..., Awaitable[T]]) -> Callable[..., T]:
         raise AssertionError
 
     def wrapped(*args, **kwargs):
-        name = threading.current_thread().name
-        inner = coro(*args, **kwargs)
         try:
-            # If a loop is currently running in this thread,
-            # use a task runner.
-            asyncio.get_running_loop()
-            if name not in _runner_map:
-                _runner_map[name] = _TaskRunner()
-            return _runner_map[name].run(inner)
+            loop = asyncio.get_running_loop()
         except RuntimeError:
-            pass
+            # Workaround for bugs.python.org/issue39529.
+            try:
+                loop = asyncio.get_event_loop_policy().get_event_loop()
+            except RuntimeError:
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
+        from jupyter_core.utils import patched_nest_asyncio
 
-        # Run the loop for this thread.
-        if name not in _loop_map:
-            _loop_map[name] = asyncio.new_event_loop()
-        loop = _loop_map[name]
-        return loop.run_until_complete(inner)
+        patched_nest_asyncio.apply(loop)
+        future = asyncio.ensure_future(coro(*args, **kwargs), loop=loop)
+        try:
+            return loop.run_until_complete(future)
+        except BaseException as e:
+            future.cancel()
+            raise e
 
     wrapped.__doc__ = coro.__doc__
     return wrapped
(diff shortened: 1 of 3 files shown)


davidbrochart:
Maybe a dumb question, but since the issue seems to be related to Tornado, has anyone tried to run Jupyter Notebook with [Jupyverse](https://github.com/jupyter-server/jupyverse)?

smacke:
> I don't think we want to go back to nest-asyncio, which has its own issues too.
Agreed (at least for the long term), the purpose of this PR is mainly to raise awareness around https://github.com/jupyter/notebook/issues/6721 -- I've found that this is usually the best way to pester more knowledgeable folks :)

> Maybe a dumb question, but since the issue seems to be related to Tornado, has anyone tried to run Jupyter Notebook with [Jupyverse](https://github.com/jupyter-server/jupyverse)?
Not sure (I haven't tried this) -- I only know that right now, for many people, if they do `pip install --upgrade jupyter` to get the latest versions of dependencies, and ran the "legacy" notebook v6 via `jupyter notebook`, they get errors like https://github.com/jupyter/notebook/issues/6721

jamescooke:
@davidbrochart I didn't know about Jupyverse (and it looks interesting) but I'm not able to run it because of the following dependency conflict (it seems that `asphalt` used by the api requires an older version of `typeguard`, [asphalt's requirement here](https://github.com/asphalt-framework/asphalt/blob/master/pyproject.toml#L33))

```
There are incompatible versions in the resolved dependencies:
  typeguard>=3.0.2 (from pandera==0.15.1->-r requirements.in (line 19))
  typeguard~=2.0 (from asphalt==4.11.1->jupyverse-api==0.1.6->jupyverse[auth,jupyterlab]==0.1.6->-r requirements.in (line 14))
```

davidbrochart:
I don't have this conflict locally. Which package depends on `pandera`?

jamescooke:
@davidbrochart Sorry I wasn't clear - (and this is off topic for this Issue probably). My large notebooks that have experienced the crashes in https://github.com/jupyter/notebook/issues/6721 use pandera to validate data at load time, it's not a dependency of anything in this repo or Jupyverse 🙇🏻 . The conflict meant that I couldn't follow up on your suggestion to try that server.

davidbrochart:
Anyway for Jupyter Notebook, a new Jupyverse plugin should be created. But if Jupyter Notebook v7 is fine for you, this will be easy since Jupyverse already has a RetroLab plugin, which should be very similar.

smacke:
That's great that we have some workarounds. However, I am concerned that the default jupyter installation (with notebook v6) is currently broken by default. E.g. if I do:

```
pip install --upgrade jupyter
jupyter notebook
```

I can reliably reproduce the issue from https://github.com/jupyter/notebook/issues/6721.

One way to reproduce is to generate a plot in a cell with matplotlib, and then hold down ctrl+enter to keep submitting the cell as fast as possible. This leaves the notebook in an unusable state.

While that particular kind of interaction may be unlikely to occur in practice, others are reporting that restart + run all followed by normal use will trigger it.

And it definitely root causes to this threaded `run_sync`.

philippjfr:
Sorry for chiming in on a PR but I didn't see any actual issue about this on here and did want to re-raise this. As you can see from https://github.com/jupyter/notebook/issues/6721 this is a major issue that is biting quite a few people. While we have put mitigation in place for our projects we still see this issue crop up quite frequently causing major issues for many [Panel](https://github.com/holoviz/panel) and [HoloViews](https://github.com/holoviz/holoviews) users. As of right now we can only recommend users pin older versions of jupyter-server and jupyter-client, which does not seem a reasonable long-term solution. Is there anything we can do to work towards a fix here?

blink1073:
I think the right solution is for notebook<7 to pin, as in this PR: https://github.com/jupyter/notebook/pull/6749.  We can't undo the changes in the stdlib that drove us to move away from `nest_asyncio`.

davidbrochart:
I recently found out about [greenletio](https://github.com/miguelgrinberg/greenletio). It looks interesting, especially the ability to use await inside a sync function:
```py
from greenletio import await_

async def async_function():
    pass

def sync_function():
    await_(async_function())
```

blink1073:
I agree, it does look promising, we're looking at using it for async PyMongo as well: https://gist.github.com/blink1073/3873afd23977a8ab94a423127f2ca2ce

davidbrochart:
The only thing is that it uses [greenlet](https://github.com/python-greenlet/greenlet), which is a C extension and probably does some low-level magic, but it works on CPython.
Not sure about PyPy though?

blink1073:
Good point, I think greenletio would need to support continuelet as well: https://doc.pypy.org/en/latest/stackless.html.

blink1073:
Ah, good, greenletio is tested on pypy: https://github.com/miguelgrinberg/greenletio/blob/db38bd2c5315cd0b9d515665f19256424fe655ee/.github/workflows/tests.yml#L25

smacke:
> We can't undo the changes in the stdlib that drove us to move away from `nest_asyncio`.

I'd be keen to learn more about this -- has anybody written a tl;dr documenting some of these issues? I came across [this extremely long thread](https://github.com/erdewit/nest_asyncio/issues/36) but find myself more confused than when I started.

blink1073:
I don't know of a good tl;dr: here's the thread I was involved in: https://github.com/python/cpython/issues/93453#issuecomment-1279979052

smacke:
> I don't know of a good tl;dr: here's the thread I was involved in: https://github.com/python/cpython/issues/93453#issuecomment-1279979052

Thanks for the pointer!

> However, I am concerned that the default jupyter installation (with notebook v6) is currently broken by default.

I happened to reread this several weeks later, and am now cringing hard. I hope folks on this thread will accept my apologies for the wording. I shouldn't have minimized the (incredible) contributions of Jupyter core contributors, especially not without context behind some of the changes being made.

davidbrochart:
Actually `greenletio` would not help us running async code from a sync function running in an event loop:
```py
import asyncio
from greenletio import await_

async def async_function():
    pass

def sync_function():
    await_(async_function())

async def main():
    sync_function()

asyncio.run(main())

# RuntimeError: await_ cannot be called from the asyncio task
```

blink1073:
That's what we plan to do for PyMongo.

blink1073:
We believe #362 fixes the issue, but in the mean time Notebook 6.5.5 has pinned its use of `jupyter_core`.  https://github.com/jupyter/notebook/issues/6721#issuecomment-1662440259.  I'm going to close this PR, since we don't want to go back to using `nest_asyncio`.

smacke:
> We believe https://github.com/jupyter/jupyter_core/pull/362 fixes the issue

> One way to reproduce is to generate a plot in a cell with matplotlib, and then hold down ctrl+enter to keep submitting the cell as fast as possible. This leaves the notebook in an unusable state.

Unfortunately, I can confirm that the above repro still reproduces the same issue on notebook v6. Here are the (potentially relevant) libraries I'm using:

jupyter_core==5.3.2
jupyter_client==8.3.1
tornado==6.3.2
notebook==6.5.6

Though, maybe it is not really an issue anymore now that notebook v7 is out.

PR #356, opened 2023-06-13
Better handling of config migration

### LINKED ISSUE #355: migrated marker may seem not to exist if os.stat not allowed
I came across this issue while running in a read-only home environment (HPC batch job) on AFS.

As far as I can tell, the issue is even documented the [Python os module reference](https://docs.python.org/3/library/os.path.html#os.path.exists)

>  On some platforms, this function may return False if permission is not granted to execute [os.stat()](https://docs.python.org/3/library/os.html#os.stat) on the requested file, even if the path physically exists.

So what happens is that here

https://github.com/jupyter/jupyter_core/blob/0ae6f9058d971489f95ea8d38b98dba26556ffa6/jupyter_core/application.py#LL171C1-L171C1

it looks like the migrated marker does not exist, so it executes `migrate()`, which that fails when it tries to write that marker.

I think the most appropriate thing would be for any operations in `migrate()` to first check if `JUPYTER_CONFIG_DIR` is actually 
 writable and if not, cancel the migration or fail back to some temp directory, like IPython does 
```python
IPython/paths.py:62: UserWarning: IPython dir '...' is not a writable location, using 
a temp directory.
```

smartass101:
better than just checking if it exists, because some on some platforms (e.g. ro AFS on HPC ) os.path.exists() may give False just because os.stat fails, even though the file actually exists

Also adds more test coveraate of the migrated marker checking

Fixes #355

### DIFF
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -170,7 +170,7 @@ class JupyterApp(Application):
         """Migrate config/data from IPython 3"""
         try:  # let's see if we can open the marker file
             # for reading and updating (writing)
-            f_marker = open(os.path.join(self.config_dir, "migrated"), 'r+')
+            f_marker = open(os.path.join(self.config_dir, "migrated"), 'r+')  # noqa
         except PermissionError:  # not readable and/or writable
             return  # so let's give up migration in such an environment
         except FileNotFoundError:  # cannot find the marker file
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -168,17 +168,17 @@ class JupyterApp(Application):
 
     def migrate_config(self):
         """Migrate config/data from IPython 3"""
-        try:# let's see if we can open the marker file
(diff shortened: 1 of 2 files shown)


jupyter_core/application.py
```diff
@@ -168,9 +168,17 @@ def ask():
...
     def migrate_config(self):
         """Migrate config/data from IPython 3"""
-        if os.path.exists(os.path.join(self.config_dir, "migrated")):
-            # already migrated
-            return
+        try:  # let's see if we can open the marker file
+            # for reading and updating (writing)
+            f_marker = open(os.path.join(self.config_dir, "migrated"), 'r+')
+        except PermissionError:  # not readable and/or writable
+            return  # so let' give up migration in such an environment
```
blink1073, line 175:
```suggestion
            return  # so let's give up migration in such an environment
```

blink1073 (approved):
Thank you!

PR #362, opened 2023-09-26
Fix event loop handling

blink1073:
Fixes https://github.com/jupyter/jupyter_client/issues/959

Fixes https://github.com/spyder-ide/spyder/issues/21338

### DIFF
--- a/jupyter_core/utils/__init__.py
+++ b/jupyter_core/utils/__init__.py
@@ -127,7 +127,6 @@ class _TaskRunner:
 
 
 _runner_map: dict[str, _TaskRunner] = {}
-_loop_map: dict[str, asyncio.AbstractEventLoop] = {}
 
 
 def run_sync(coro: Callable[..., Awaitable[T]]) -> Callable[..., T]:
@@ -161,9 +160,11 @@ def run_sync(coro: Callable[..., Awaitable[T]]) -> Callable[..., T]:
             pass
 
         # Run the loop for this thread.
-        if name not in _loop_map:
-            _loop_map[name] = asyncio.new_event_loop()
-        loop = _loop_map[name]
+        try:
+            loop = asyncio.get_event_loop()
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
         return loop.run_until_complete(inner)
 
     wrapped.__doc__ = coro.__doc__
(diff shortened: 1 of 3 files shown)


blink1073:
cc @maartenbreddels this was a nice typing cleanup.   :)

blink1073:
I'll make the release tomorrow.

maartenbreddels:
Nice indeed :)

samuelan:
I have some codes that use run_sync(coroutine_a), and with jupyter_core==5.3.2, it stops working. The code complains 
 File "/opt/conda/lib/python3.9/site-packages/jupyter_core/utils/__init__.py", line 168, in wrapped
    return loop.run_until_complete(inner)
  File "/opt/conda/lib/python3.9/asyncio/base_events.py", line 623, in run_until_complete
    self._check_running()
  File "/opt/conda/lib/python3.9/asyncio/base_events.py", line 583, in _check_running
    raise RuntimeError('This event loop is already running')
RuntimeError: This event loop is already running

once I revert to 5.3.1 version. It started to work. Do codes that depend on jupyter_core need to change after this change, any documentation?

blink1073:
Hi @samuelan, can you please open a new issue with some code that reproduces the error?

samuelan:
> Hi @samuelan, can you please open a new issue with some code that reproduces the error?

Sure. It's a bit involved. I will try a repro with simplified codes, and circle back when I get it.

jupyter_core/utils/__init__.py
```diff
@@ -161,9 +160,11 @@ def wrapped(*args: Any, **kwargs: Any) -> Any:
             pass
 
         # Run the loop for this thread.
-        if name not in _loop_map:
-            _loop_map[name] = asyncio.new_event_loop()
-        loop = _loop_map[name]
+        try:
+            loop = asyncio.get_event_loop()
+        except RuntimeError:
```
tornaria, line 165: Do you actually want `loop = asyncio.get_running_loop()` here?

According to https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop this is deprecated in python 3.12 and will emit `https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop`. This causes some test failures (e.g. in jupyter_client).

tornaria: What I suggested above does fix the test failures in jupyter_client I was getting, at the expense of causing other test failures in jupyter_client.

The only thing that makes the jupyter_client testsuite pass (on python 3.12) is reverting the current PR. I'll downgrade to jupyter_core 5.3.1 for the time being.

blink1073: Thanks for the heads up @tornaria, I opened #367

PR #364, opened 2023-10-02
Lowercase APPNAME on macOS Homebrew

singingwolfboy:
The [latest release of `platformdirs`](https://github.com/platformdirs/platformdirs/releases/tag/3.11.0) contains support for Homebrew-installed software, under the `/opt/homebrew` directory on macOS. In order for Jupyter to support this, we should also check to see if the Jupyter software is installed/managed by Homebrew on macOS, and if so, treat the `APPNAME` the same way we do on Linux: lowercase, rather than capitalized.


defjaf:
Worth changing, but at least by default the filesystem is case *preserving*, but not case *sensitive*, in which case this is more of an aesthetic issue. (But it is at least possible to have case sensitive macOS filesystems.)

blink1073 (approved):
Thank you!

PR #382, opened 2023-12-26
Use supported event loop behavior

blink1073:
Based on discussion in https://github.com/python/cpython/issues/100160, we should not be relying on `asyncio.get_event_loop`, since they are in fact going to make it an alias for `asyncio.get_running_loop`.

As of #381, we are starting the event loop before instantiating our apps, so we should not see any more issues like https://github.com/jupyter/notebook/issues/6721.

Based on results in https://github.com/ipython/ipykernel/pull/1184, we'll want to normally use the default event loop on Windows, unless we know we're running a tornado app.  This PR adds a `prefer_selector_loop` option to `get_event_loop` so we can override `_prefer_selector_loop` in `jupyter_server` apps to prefer the selector loop.

### DIFF
--- a/jupyter_core/application.py
+++ b/jupyter_core/application.py
@@ -76,6 +76,9 @@ class JupyterApp(Application):
     aliases = base_aliases
     flags = base_flags
 
+    # Set to True for tornado-based apps.
+    _prefer_selector_loop = False
+
     def _log_level_default(self) -> int:
         return logging.INFO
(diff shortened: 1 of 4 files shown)


blink1073:
I'm going to close this in light of #385.  After we fix #384 and release, I'll make a new PR that ensures we are handling the problems raised in #383.

PR #386, opened 2023-12-31
Fix export of package version

### LINKED ISSUE #384: `__version__` attribute no longer available
It seems like `__version__` has been removed from the main namespace in jupyter_core 3.6. I have noted that this had led to issues in at least two other applications that rely on being able to read the version number from `__version__`, see https://github.com/conda-forge/quarto-feedstock/issues/34#issuecomment-1871421353 and https://gitlab.com/joelostblom/session_info/-/issues/14 for details. Would it be possible to reintroduce the `__version__` attribute (or at least mark it as deprecated for a while), or is there a reason for not using it any longer? From my understanding it is more common to have `__version__` as an attribute on the main module rather than in a `version.__version__` submodule, but there might very well be good reasons for the latter that I am not aware of.

I believe it happened in this PR https://github.com/jupyter/jupyter_core/pull/376 so cc @blink1073.

blink1073:
Fixes #384

### DIFF
--- a/jupyter_core/__init__.py
+++ b/jupyter_core/__init__.py
@@ -1,0 +1,3 @@
+from __future__ import annotations
+
+from .version import __version__, version_info  # noqa: F401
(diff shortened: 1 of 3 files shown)


kevin-bates (approved):
LGTM, thanks @blink1073!

Issue #383, opened 2023-12-27
0.5.6 causes "This event loop is already running" Error


a3626a:
This new 5.6 version causes this error : (https://github.com/jupyter/jupyter_core/pull/381)
```
notebook Traceback (most recent call last):
notebook   File "/usr/local/bin/jupyter-labhub", line 8, in <module>
notebook     sys.exit(main())
notebook   File "/usr/local/lib/python3.10/site-packages/jupyterlab/labhubapp.py", line 41, in main
notebook     return SingleUserLabApp.launch_instance(argv)
notebook   File "/usr/local/lib/python3.10/site-packages/jupyter_core/application.py", line 307, in launch_instance
notebook     loop.run_until_complete(coro)
notebook   File "/usr/local/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
notebook     return future.result()
notebook   File "/usr/local/lib/python3.10/site-packages/jupyter_core/application.py", line 292, in _async_launch_instance
notebook     app.initialize(argv)
notebook   File "/usr/local/lib/python3.10/site-packages/jupyterhub/singleuser/mixins.py", line 972, in initialize
notebook     result = super().initialize(*args, **kwargs)
notebook   File "/usr/local/lib/python3.10/site-packages/jupyterhub/singleuser/mixins.py", line 643, in initialize
notebook     super().initialize(argv)
notebook   File "/usr/local/lib/python3.10/site-packages/traitlets/config/application.py", line 118, in inner
notebook     return method(app, *args, **kwargs)
notebook   File "/usr/local/lib/python3.10/site-packages/jupyter_server/serverapp.py", line 2610, in initialize
notebook     self.init_httpserver()
notebook   File "/usr/local/lib/python3.10/site-packages/jupyterhub/singleuser/mixins.py", line 668, in init_httpserver
notebook     self.io_loop.run_sync(super().init_httpserver)
notebook   File "/usr/local/lib/python3.10/site-packages/tornado/ioloop.py", line 533, in run_sync
notebook     self.start()
notebook   File "/usr/local/lib/python3.10/site-packages/tornado/platform/asyncio.py", line 205, in start
notebook     self.asyncio_loop.run_forever()
notebook   File "/usr/local/lib/python3.10/asyncio/base_events.py", line 592, in run_forever
notebook     self._check_running()
notebook   File "/usr/local/lib/python3.10/asyncio/base_events.py", line 584, in _check_running
[...]

PR #387, opened 2023-12-31
Modernize event loop behavior

blink1073:
Based on discussion in https://github.com/python/cpython/issues/100160, we should not be relying on `asyncio.get_event_loop`, since they are in fact going to make it an alias for `asyncio.get_running_loop`.

This ensures we install an event loop before launching our apps, so we should not see any more issues like https://github.com/jupyter/notebook/issues/6721.

Additionally, it adds a new class, `JupyterAsyncApp` that runs `initialize_async()` and `start_async()` in an event loop.

Based on results in https://github.com/ipython/ipykernel/pull/1184, we'll want to normally use the default event loop on Windows, unless we know we're running a tornado app.  This PR adds a `prefer_selector_loop` option to `ensure_event_loop` so we can override `_prefer_selector_loop` in `jupyter_server` apps to prefer the selector loop.

Based on failures seen in #383, we add tests to ensure we are not interfering with existing event loop behavior.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -79,7 +79,7 @@ jobs:
         uses: jupyterlab/maintainer-tools/.github/actions/downstream-test@v1
         with:
           package_name: pytest_jupyter
-          package_spec: pip install -e ".[client,server]"
+          package_spec: pip install -e ".[test,client,server]"
 
   downstreams_check: # This job does nothing and is only used for the branch protection
     if: always()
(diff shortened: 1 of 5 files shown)


gvanrossum (commented):
Could you name it something other than ‘get_event_loop’, to avoid confusion with the asyncio function by that name? Maybe ‘ensure_event_loop’?

blink1073:
> Could you name it something other than ‘get_event_loop’, to avoid confusion with the asyncio function by that name? Maybe ‘ensure_event_loop’?

Done, thank you!

PR #406, opened 2024-10-09
Remove option from extension that is not present anymore.

Carreau:
This was used by sphinxcontrib_github_alt, that is not in the list of extensions.


Carreau:
Sorry, this is a sphinx option now.

PR #412, opened 2025-02-26
A bunch of pre-commit and CI upgrade.

Carreau:
- precommit-ci is not sending any updates because of branch protection rules.
- Update and fix corresponding errors.
- run mypy as part of CI.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ ci:
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v5.0.0
     hooks:
       - id: check-case-conflict
       - id: check-ast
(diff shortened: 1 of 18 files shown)


.pre-commit-config.yaml
```diff
@@ -39,22 +39,22 @@ repos:
...
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: "v1.8.0"
+    rev: "v1.15.0"
     hooks:
       - id: mypy
         files: jupyter_core
         stages: [manual]
-        args: ["--install-types", "--non-interactive"]
-        additional_dependencies: ["traitlets>=5.13", "platformdirs>=3.11"]
```
Carreau, line 54: `"--install-types", "--non-interactive"` will actually silence some errors are they are ment to be run locally, I  think this is a mypy bug.

PR #429, opened 2025-05-26
validate inputs to is_hidden

### LINKED ISSUE #423: paths.is_hidden() Erroneously Reports Files as Hidden
A bug in a previous version of Jupyter led to errors when trying to rename a file.
After a bit of digging, I think I've tracked down the root cause to
[line 497 of jupyter_core/paths.py](https://github.com/jupyter/jupyter_core/blob/283b128de9d3e4dabf2a23f971d839a4d468b961/jupyter_core/paths.py#L497),
where the assignment `inside_root = abs_path[len(abs_root) :]` assumes that the abs_path will have abs_root as a prefix.
In some cases, the abs_path could get trimmed down to exactly the file extension,
like in this case:

```python
is_hidden("filename.ipynb", "/home/aa")
```

where `filename` is exactly the same length as `/home/aa`.
The resulting trimmed file extension `.ipynb` looks like a hidden file when checked.

A simple way to fix this would be to check that abs_root is a prefix of abs_path before trimming:

```python
if abs_path.startswith(abs_root):
    inside_root = abs_path[len(abs_root):]
else:
    inside_root = abs_path
```

I'd also propose to fix the check on abs_root on line 495,
since the call to `os.path.normpath()` on line 487 ensures it will never be an empty string:

```python
    abs_path = os.path.normpath(abs_path)
    if abs_root == "":
        abs_root = abs_path.split(os.sep, 1)[0] + os.sep
    else:
        abs_root = os.path.normpath(abs_root)
```

minrk:
Both path inputs to `is_hidden()` must be absolute, as is documented. But this wasn't checked, which could lead to incorrect results for invalid inputs. These cases are now ValueErrors instead of undefined behavior.

closes #423

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -550,7 +550,7 @@ def is_hidden(abs_path: str | Path, abs_root: str | Path = "") -> bool:
     if abs_root:
         abs_root = Path(os.path.normpath(abs_root))
     else:
-        abs_root = abs_path.root
+        abs_root = Path(abs_path.root)
 
     if abs_path == abs_root:
         # root itself is never hidden
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -550,7 +550,7 @@ def is_hidden(abs_path: str | Path, abs_root: str | Path = "") -> bool:
     if abs_root:
         abs_root = Path(os.path.normpath(abs_root))
     else:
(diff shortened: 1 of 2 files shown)


Carreau:
(rebased)

Carreau:
Failure unrelated, merging.

PR #430, opened 2025-05-26
use env config path for system config path when no system config path

minrk:
SYSTEM_CONFIG_PATH hasn't technically been guaranteed to be non-empty, but jupyter-server has assumed it is: https://github.com/jupyter-server/jupyter_server/issues/1527 and in practice, it ~always has been until 5.8 because the empty case was weird and rare.

We already use sys.prefix for SYSTEM_JUPYTER_PATH in a similar situation, so it's probably simplest to do the same in config path.

cc @krassowski

I can do a 5.8.1 with this.

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -411,7 +411,13 @@ def jupyter_config_path() -> list[str]:
             if userdir not in user:
                 user.append(userdir)
 
-    env = [p for p in ENV_CONFIG_PATH if p not in SYSTEM_CONFIG_PATH]
+    # Windows usually doesn't have a 'system' prefix,
+    # so 'system' and 'env' are the same
+    # make sure that env can still be preferred in this case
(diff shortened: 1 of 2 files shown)


krassowski:
I'm not sure if this will not have any unintended side-effects, but it sounds like it's a better situation than having jupyter server extension not work at all.

> We already use sys.prefix for SYSTEM_JUPYTER_PATH in a similar situation, so it's probably simplest to do the same in config path.

I could not find that maybe I am just too tired), can you leave a reference for posterity?

minrk:
I looked around, and https://github.com/jupyter-server/jupyter_server/issues/1527 is the only reference to SYSTEM_CONFIG_PATH I can find that would be sensitive to this. All readers of this path rely on `jupyter_config_path` which handles this and avoids duplicate entries (with a test verifying it), so I think this should be safe.

PR #431, opened 2025-05-26
Convert license settings into PEP-639 format

jbedorf:
The change [here](https://github.com/jupyter/jupyter_core/commit/abc66f1a4418d5aa1f0b7ec56b1e03e135d63a14) led to tools that extract the license (like `pip-licenses`) to fail, because the content of the `LICENSE` file became the license identifier.

This change converts the settings into the PEP-639 format and allows tools to correctly identify the license of installed packages. 

The BSD-3-Clause format is the one suggested [here](https://spdx.org/licenses/BSD-3-Clause.html).


minrk:
Thanks!

PR #433, opened 2025-05-28
Fix missing `f` specifier in f-string used to print config file path

krassowski:
Fixes https://github.com/jupyterlab/jupyterlab/issues/17601


danyeaw (approved):
Thanks @krassowski!

Issue #309, opened 2022-11-09
6.0: Change Default for JUPYTER_PLATFORM_DIRS to "true"


blink1073:

minrk:
We'll need to make sure to migrate user config files automatically to the new locations, to avoid major disruptions to users. If we don't migrate files automatically, we should at least check, and provide a one-line migration command from the old location to the new.

We still have [the migration code](https://github.com/jupyter/jupyter_core/blob/main/jupyter_core/migrate.py) for the move from .ipython to .jupyter, and should update it to handle the soon-to-be deprecated Jupyter paths, too.

minrk:
Also related to this: the platform dirs change appears to still be entirely undocumented outside jupyter-core's changelog.

It should at least be covered [here](https://docs.jupyter.org/en/latest/use/jupyter-directories.html) and somewhere in the jupyter-core docs before the deprecation clock can safely start.

minrk:
#447 implements @ivanov's proposal to remove the deprecation warning.

PR #445, opened 2025-10-08
start to test on 3.14 and 3.14-free-threaded

Carreau:


Carreau:
@minrk I see you recently removed and readded Python 3.8, would you be or to drop 3.8, and bump version number of 6.0 ?

Carreau:
And I don't know what to do with the fialures, `asyncio.set_event_loop_policy` is anyway deprecated and will be removed in 3.16.

minrk:
That was to get a security fix out without changing supported Python versions, fine to drop 3.8 and 3.9 now.

minrk:
For the event loop policy, we could consider just removing it. pyzmq now works with proactor if tornado is available (via an additional thread), so if we require tornado 6.1, all should be okay (if less efficient than selector).

We could also just ignore the deprecation warning. The code will still work (i.e. it will do nothing) after policies are removed, because they are protected by a check for an ImportError.

I'm still incredibly frustrated that asyncio folks decided proactor was ready to be the default when it still lacks basic functionality of all other loops that needs to be worked around with threads.

Carreau:
Rhaa mypy bug [seem to be known](https://github.com/python/mypy/issues/10773).

minrk:
my inclination: suppress the deprecation warning for now, then we can propose a new way to do it without policies (I think an `init_ioloop` that instantiates a loop directly will work) in its own PR.

Carreau:
I think I'll be in favor of ignoring the warnings only python 3.14, and revisit for 3.15

jupyter_core/utils/__init__.py
```diff
@@ -182,6 +177,17 @@ def ensure_event_loop(prefer_selector_loop: bool = False) -> asyncio.AbstractEve
         loop = asyncio.get_running_loop()
     except RuntimeError:
         if sys.platform == "win32" and prefer_selector_loop:
+            if (3, 14) <= sys.version_info < (3, 15):
+                # ignore deprecation only for 3.14 and revisit later.
+                with warnings.catch_warnings():
+                    warnings.simplefilter(
```
minrk: I think you want

```suggestion
                    warnings.filterwarnings(
```

for the `message` argument

and maybe just filter on `.*WindowsSelectorEventLoopPolicy.*` to avoid missing irrelevant changes in language, or if removal happens to be kicked down the road to 3.17, etc.

jupyter_core/utils/__init__.py
```diff
@@ -182,6 +177,17 @@ def ensure_event_loop(prefer_selector_loop: bool = False) -> asyncio.AbstractEve
...
+                        "ignore",
+                        DeprecationWarning,
+                        message=".*'asyncio.WindowsSelectorEventLoopPolicy' is deprecated and slated for removal in Python 3.16.*",
+                    )
+                    loop = asyncio.WindowsSelectorEventLoopPolicy().new_event_loop()
+            else:
+                loop = asyncio.WindowsSelectorEventLoopPolicy().new_event_loop()
             loop = asyncio.WindowsSelectorEventLoopPolicy().new_event_loop()
```
minrk: This line is now extra, creating two loops and only using one

Carreau:
All passing, and please don't squash I properly cleaned-up each commit.

Carreau:
Current version is 5.8.1, once this is merged, I suggest we bump version, and release soon. I'll suggest 2 new possible version numbers in two separate comments. Please upvote your peference.

Carreau:
5.9, nothing specific but did drop 3.10 (and 3.9) and there was no release since 3.9 drops.

Carreau:
6.0, this will let us do some cleanups in particular this deprecation warning:

```
else:
    deprecation(
        "Jupyter is migrating its paths to use standard platformdirs\n"
        "given by the platformdirs library.  To remove this warning and\n"
        "see the appropriate new directories, set the environment variable\n"
        "`JUPYTER_PLATFORM_DIRS=1` and then run `jupyter --paths`.\n"
        "The use of platformdirs will be the default in `jupyter_core` v6"
    )
```

EDIT: version published Nov 9 2022

Carreau:
We can also do 5.9, and do 6.0 in a month or so with all the cleanup we want.

minrk:
I think we can do 5.9 with these changes, I don't have a problem with dropping Python versions on minor releases, the tooling handles that well enough.

I haven't pulled the trigger on completing the platformdirs change because I think it's not complete enough - https://github.com/jupyter/jupyter_core/issues/309 I don't think we should _start_ the deprecation clock until we have had migration in place for quite some time, which we still don't. At this point, I agree with @ivanov that we should remove the deprecation warning and stop telling people the default is going to change.

minrk:
#447 removes the deprecation warning

Carreau:
Thanks !

minrk:
draft release: https://github.com/jupyter/jupyter_core/releases/tag/untagged-9833f440530aecf4ad1b ready to go?

jupyter_core/utils/__init__.py
```diff
@@ -182,7 +177,17 @@ def ensure_event_loop(prefer_selector_loop: bool = False) -> asyncio.AbstractEve
         loop = asyncio.get_running_loop()
     except RuntimeError:
         if sys.platform == "win32" and prefer_selector_loop:
-            loop = asyncio.WindowsSelectorEventLoopPolicy().new_event_loop()
+            if (3, 14) <= sys.version_info < (3, 15):
+                # ignore deprecation only for 3.14 and revisit later.
+                with warnings.catch_warnings():
+                    warnings.filterwarnings(
+                        "ignore",
+                        DeprecationWarning,
```
Darshan808, line 185:
```suggestion
                        category=DeprecationWarning,
```

This is causing `TypeError` in downstream package `jupyterlab_server` when testing on Python 3.14 on Windows.
CC: @Carreau

Carreau:
5.9 is already out.

jupyter_core/utils/__init__.py
Darshan808: Opened https://github.com/jupyter/jupyter_core/pull/449

PR #447, opened 2025-10-13
remove deprecation warning, platformdirs won't become default

### LINKED ISSUE #309: 6.0: Change Default for JUPYTER_PLATFORM_DIRS to "true"

minrk:
Back off plan to deprecate longstanding default paths. Folks can still opt-in to platformdirs, but there is little reason to change the defaults. Changing defaults is necessarily a long and sensitive process, and needs to come with documentation, support for migration, etc., none of which has happened in the long time since the proposed deprecation started.

Simpler approach: keep defaults, stop warning about them.

closes #309

### DIFF
--- a/jupyter_core/paths.py
+++ b/jupyter_core/paths.py
@@ -22,8 +22,6 @@ from typing import Any, overload
 
 import platformdirs
 
-from .utils import deprecation
-
 pjoin = os.path.join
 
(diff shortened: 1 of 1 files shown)


Carreau (commented):
Feel free to self-merge, I'll leave it a bit to get other a chance to chime in.

ivanov (approved):

PR #449, opened 2025-10-16
Fix `filterwarnings()` TypeError

Darshan808:
This `filterwarnings()` call is causing `TypeError: filterwarnings() got multiple values for argument 'message'` in downstream package `jupyterlab_server` when testing on Python 3.14 and jupyer_core 5.9 on windows.


ivanov:
Thanks for the PR, allowing worflows to run.

minrk:
Thanks! I'll cut 5.9.1 shortly

minrk:
5.9.1 on its way out now

